PR #14, opened 2020-05-12
Transaction

### LINKED ISSUE #5: [Transaction] Serialization
The `Transaction` type must export a `Serialize` method that encodes all transaction information in a byte slice.

An internal `serialize` method, instead, must be implemented in order to support the different type of serialization, i.e. the serialization of the tx for producing a signature and the serialization including/excluding witnesses.

### LINKED ISSUE #6: [Transaction] Deserialization
The `transaction` package must export a `NewTxFromBuffer` method that deserializes the given raw transaction in bytes format, returning a `*Transaction` instance or an error.

Also, the package should export a `NewTxFromString`, that parses the hex-encoded transaction into raw bytes and calls the method defined above.

### LINKED ISSUE #7: [Transaction] Copy
The `Transaction` type must export a `Copy` method that returns a copies and returns a new instance of the transaction.

altafan:
## Transaction

Changes to type definition:

* `Hash` has been changed from `chainhash.Hash` to `[]byte` because when serializing the only way to get the bytes from the chainhash.Hash type is to slice it with `[:]` that apparently reverses the hash introducing an error in the serialization.
* `Issuance` becomes a pointer to `TxIssuance` (`*TxInssuance`)
* `Witness` and `PeginWitness` becomes of type `TxWitness` that is just a named `[][]byte` but with a method `SerializeSize` useful when serializing the transaction

Exported methods:
* `TxHash()` and `WitnessHash()` to calculate the transaction hash and its witness hash (if the tx does not contain witness this returns just the transaction hash)
* `Weight()` and `VirtualSize()` to calculate the size of the transaction
* `Copy()` to make a deep copy of the current transaction
* `Serialize()` to serialize the transaction into bytes
* `ToHex()` to serialize the transaction and encode into hex string format
* `SerializeSize(allowWitness, forSignature bool)` to get the length of a specific serialization
* `NewTxFromBuffer` and `NewTxFromHex` to deserialize given transaction whether in bytes or string format.

## Tx input and output

* This adds a `NewTxInput` and a `NewTxOutput` that return an unsigned input and an output with the given parameters
* Both types now expose a method `SerializeSize`, similar to the `TxWitness` one, that help when serializing the transaction

This closes #5.
This closes #6.
This closes #7.

Please @tiero, review this.


transaction/transaction.go
```diff
@@ -73,9 +314,249 @@ func (tx *Transaction) TxHash() chainhash.Hash {
...
+	return size
+}
+
+// Serialize returns the serialization of the transaction.
+func (tx *Transaction) Serialize() ([]byte, error) {
+	return tx.serialize(nil, true, false, false)
+}
+
+// String returns the serializarion of the transaction in hex enncoding format.
+func (tx *Transaction) String() (string, error) {
```
tiero, line 438: Would be Better calling `ToHex` ?

Issue #9, opened 2020-05-12
[Payment] FromPayment


altafan:
A `FromPayment` function similar to the existing `FromPublicKey` one must be added to the `payment` package in order to return a `Payment` from another `*Payment` struct and a network.

Ref: #8.

tiero:
Can you elaborate better on what you expect to be a Redeem script? What's the difference then from #10 ?

altafan:
The `Redeem` type is used for nested script hash addresses, for example for a p2sh-p2wpkh or p2sh-p2wsh.

The `FromRedeem` function would set the `Hash` field by computing the `hash160` of `Redeem.Hash`.

In the light of these considerations, I'm re-evaluating the usefulness of `Script` and now thinking that it would just become a copy of the `Hash` field and it would be better to drop it.

altafan:
`Script`, instead, is needed for non-nested p2sh addresses, like multisigs:
from this field, that in our example would contain the script formed by the _n_ public keys and the _m_ threshold, we would derive the `Hash`, as always, by computing the `hash160`.

`FromScript` let the user to create his own script and pass it to `Payment` to get the p2sh or p2wsh address.

We would likely implement also a `FromPubkeys` or whatever method to facilitate the creation of a multisig address.

tiero:
@altafan Please make some example in pseudo code of what is needed, some examples of inputs and expected outputs.

altafan:
p2sh-p2wpkh example:

```
// create a p2wpkh first
p2wpkh := FromPublicKey(myPubkey)
// now nest the payment into a p2sh
p2sh := FromRedeem(p2wpkh)
// finally get the address
address := sh.ScriptHash()
```

In the first step, we simply generate a Payment from a public key.
Now, instead, of deriving the base58/bech32 address from the payment's `Hash`, we want to wrap it into another payment.
So, in the second step, we make use of `FromRedeem` to nest the p2wpkh into a p2sh.
This technically means taking the hash of the given payment and hashing it again to produce the final hash that will be encoded, in this case, in base58 format.

@sekulicd please, ask if it's still unclear.

tiero:
@sekulicd @altafan  I edited the description of the task because actually it's NOT a **redeem script** what we need to pass as a parameter, but another Payment object. For that, we have the `FromScript` where (I suppose) we would need to pass the reddems script as array of bytes or hex string.

altafan:
Good point. I think it's better renaming.

[...]

PR #21, opened 2020-05-13
ScriptHash() method added to *Payment type.

### LINKED ISSUE #11: [Payment] ScriptHash
A `ScriptHash` method similar to the `PubKeyHash` one must be added to the `*Payment` type in order to return a base58check encoded p2sh address string generated from `Hash`.

Ref: #8.

sekulicd:
ScriptHash() method added to *Payment type.
Method returns base58 P2SH address.

It closes #11

### DIFF
--- a/payment/payment_test.go
+++ b/payment/payment_test.go
@@ -29,3 +29,12 @@ func TestSegwitAddress(t *testing.T) {
 		t.Errorf("TestSegwitAddress: error when encoding segwit")
 	}
 }
+
+func TestScriptHash(t *testing.T) {
+	_, publicKey := btcec.PrivKeyFromBytes(btcec.S256(), privateKeyBytes)
+
+	pay := payment.FromPublicKey(publicKey, &network.Regtest)
+	if pay.ScriptHash() != "Xa9r81eGyEiNZpMXizeNvk2c8BkGG7K3UQ" {
+		t.Errorf("TestScriptHash: error when encoding script hash")
+	}
+}
--- a/payment/payment.go
+++ b/payment/payment.go
@@ -49,6 +49,13 @@ func (p *Payment) WitnessPubKeyHash() string {
 	return addr
 }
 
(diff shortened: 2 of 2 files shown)


payment/payment_test.go
```diff
@@ -29,3 +29,12 @@ func TestSegwitAddress(t *testing.T) {
 		t.Errorf("TestSegwitAddress: error when encoding segwit")
 	}
 }
+
+func TestScriptHash(t *testing.T) {
+	_, publicKey := btcec.PrivKeyFromBytes(btcec.S256(), privateKeyBytes)
+
+	pay := payment.FromPublicKey(publicKey, &network.Regtest)
```
tiero, line 36: This is not semantically correct. The Hash, in this case, should not represent the hash of a supplied public key, but a redeem script. https://en.bitcoin.it/wiki/BIP_0016

The correct way to do this would be to implement first a FromRedeem factory function, where you could pass then a PubKeyHash or WitnessPubKeyHash

tiero: Something like 

```go
pkh := payment.FromPublicKey(publicKey,Regtest)
psh := payment.FromRedeem(pkh,Regtest)
address := psh.ScriptHash()
```

tiero: ScriptHash should use the `Hash` field, that should be valorized by the FromReedem function using the `Script`

sekulicd: This task #11, is dependent on task #9 cause ScriptHash is invalid without the creation of Payment type using Factory func FromPayment(previously named FromRedeem).
Should I add changes that were meant to be in task #9 in this PR?
@tiero please advise

tiero: @sekulicd  Yes it's correct. Better way is to close this PR, open a new one for task #9 and only then reopen a new one for task #11. Thanks

@altafan For the future: this task should have been created and/or assigned only AFTER the task #9 got merged in master, in order to not have dependency with other tasks.

PR #24, opened 2020-05-15
[Pset] support for (de)-serialization 

### LINKED ISSUE #17: [PSET] Basic support for serialization/deserialization
Our implementation of PSETs (Partially Signed Elements Transaction) will reuse the one of [btctutil](https://github.com/btcsuite/btcutil/tree/master/psbt) as much as possible, as already sketched with the existing commits.

Taking a look at btctutil, the very first step for supporting PSET is to import the files `partial_input.go` and `partial_output.go` and all those they depend on (these other files include other data structures, constants and helper functions). 

These 2 files that I would rename `pset_input.go` and `pset_output.go`, along with the existing `pset.go` (that, of course, needs to be updated since it now contains just "placeholder" methods) essentially handle the creation, the validation, and the serialization of the data structure representing a partial transaction, its inputs, and its outputs.

### LINKED ISSUE #18: [PSET Input] Create
A method `NewPsetInput` accepting either a witness or non witness utxo and returning a `*PInput` must be added to the package in order to be able to create inputs.

Ref #17.

### LINKED ISSUE #19: [PSET Input] (De)Serialize
A `serialize` method accepting a `io.Writer` where to write the serialized inpu must be added to the package. 
The same applies for the `deserialize` method that deserialize the raw inputs from the given `io.Reader`.

Ref #17.

### LINKED ISSUE #20: [PSET Input] Validate
A method `IsSane()` returning a bool value, that verifies if an input is valid, must be added to the package.

### LINKED ISSUE #22: [PSET Output] Create
The btcutil's implementation of a Bitcoin partial output does not need any changes to be valid also for the Elements version, thus it should be enough to just rewrite it in our repo.

### LINKED ISSUE #23: [Pset] Elements support
Starting from the btcutil's implementation of a Bitcoin partial transaction, the mandatory change to do here is to fix the encoding prefix (usually defined [`magic`](https://github.com/btcsuite/btcutil/blob/master/psbt/psbt.go#L26))  since it changes for the Elements version from `psbt` (`[0x70, 0x73, 0x62, 0x74]` in bytes) to `pset` (`[0x70, 0x73, 0x65, 0x74]` in bytes).

The serialization/deserialization implementation does not need any changes, but I would change the methods exported by the `*Pset` type.

Regarding the deserialization, for the sake of usability, we should export 2 new methods `NewPsetFromHex` and `NewPsetFromBase64` in favor of [`NewFromRawBuffer`](https://github.com/btcsuite/btcutil/blob/master/psbt/psbt.go#L177).
They would be in charge of decoding the hex/base64 string into a byte slice and then calling the `NewFromRawBuffer`, that I would rename `deserialize` (so we stop exporting it).

Similarly, for serializing a partial transaction, I would rename the [`B64Encode`](https://github.com/btcsuite/btcutil/blob/master/psbt/psbt.go#L370) method to `ToBase64`, add a `ToHex` method that serializes in hex encoding and rename `Serialize` to `serialize` so we avoid exporting a low-level function.

altafan:
Before this the pset package was just defining the type `Pset` and some methods that were not actually implemented.

This introduces all those types and methods necessary to be able to serialize/deserialize a partial transaction.

### Input
```
type PInput struct {
	NonWitnessUtxo     *transaction.Transaction
	WitnessUtxo        *transaction.TxOutput
	PartialSigs        []*PartialSig
	SighashType        txscript.SigHashType
	RedeemScript       []byte
	WitnessScript      []byte
	Bip32Derivation    []*Bip32Derivation
	FinalScriptSig     []byte
	FinalScriptWitness []byte
	Unknowns           []*Unknown
}
```
This is the only type which implementation slightly differs from the btcutil one since we had to change the definition of `WitnessUtxo`.
This field represents a transaction output, so it does not just contain a script and a value (raw bytes, not a number here), but also an asset and a nonce.
For this reason, the existing btcutil's `readTxOut` has been changed accordingly and a new `writeTxOut` has been created to help in the serialization of a witness utxo.

### Output

```

type POutput struct {
	RedeemScript    []byte
	WitnessScript   []byte
	Bip32Derivation []*Bip32Derivation
}
```

Nothing relevant to say here since we were able to totally reuse the btcutil's implementation.

### Pset

```
type Pset struct {
	UnsignedTx *transaction.Transaction
	Inputs []PInput
	Outputs []POutput
	Unknowns []Unknown
}
```

The main difference of our partial transaction representation compared to the btcutil one resides exporting different methods, but the implementation itself of serialization/deserialization process has not changed.

For serializing a partial transaction, our pset provides 2 methods `ToBase64` and `ToHex` that return the serialized transaction in the proper encoding format.
For comparison, the btcutil version exposes `B64Encode` for serializing in base64 encoding and  `Serialize` that returns the raw bytes.

For deserializing a partial transaction, our pset exposes 3 methods `NewPsetFromBase64`, `NewPsetFromHex` and `NewPsetFromUnsignedTx`. It is possbile, thus, to deserialize a partial transaction from a base64/hex encoding format and also from a `Transaction` data structure.
For comparison, the btcutil version provides `NewFromUnsignedTx` to parse a `Transaction` instance, and `NewFromRawBytes` to parse the raw bytes of a serialized psbt.

This closes #17.
This closes #18.
This closes #19.
This closes #20.
This closes #22.
This closes #23.

Please @tiero, review this.


tiero:
@altafan So basically we introduce Hex encoding, something btcutil didn'have and we drop the raw bytes, that's correct?

tiero:
If you are resuing a file without touching from `btcutil` you should import and not copy paste here. Are all the files being touched? Because it's difficult to see a diff betwenne the two

altafan:
Correct. I want to specify that what you mean with "drop" here actually means "make the method unexported".

altafan:
>If you are resuing a file without touching from btcutil you should import and not copy paste here. Are all the files being touched? Because it's difficult to see a diff betwenne the two

Yes good point. I'll make a commit importing from btcutil when possible and removing all not needed files.

altafan:
We now import all the constants (mainly error messages) from btcutil rather than redefining them.

We also import the types `Bip32Derivation` and `PartialSig`, but we needed to rewrite some btcutil helper functions that are not exported from the package. All these methods can be find into `utils.go`.

Even if the implementation of partial output is not affected from changes, we still need to reimplement it since the version of btcutil does not export the serialization/deserialization methods.

tiero:
@altafan also types seems to me identical

PR #25, opened 2020-05-15
FromPayment method added

### LINKED ISSUE #9: [Payment] FromPayment

sekulicd:
Before this, there was no option to create Payment based on another Payment.

This introduces a new Factory method for the creation of a Payment based on another Payment. Internally method is setting hash, of a new Payment, by taking hash from the input and re-hash it.

This closes #9 

Please, @altafan, review this.

### DIFF
--- a/payment/payment.go
+++ b/payment/payment.go
@@ -32,7 +32,7 @@ func FromPublicKey(pubkey *btcec.PublicKey, network *network.Network) Payment {
 
 // FromPayment creates a Payment struct from a another Payment
 func FromPayment(payment *Payment, network *network.Network) Payment {
-	buf := payment.Hash
+	buf := payment.Script
 	hash := hash160(buf)
 	return Payment{network, payment.PublicKey, hash, nil, nil, nil}
 }
--- a/payment/payment.go
+++ b/payment/payment.go
@@ -30,6 +30,13 @@ func FromPublicKey(pubkey *btcec.PublicKey, network *network.Network) Payment {
 	return Payment{network, pubkey, hash, nil, nil, nil}
 }
 
+// FromPayment creates a Payment struct from a another Payment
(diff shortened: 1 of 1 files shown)


payment/payment.go
```diff
@@ -30,6 +30,13 @@ func FromPublicKey(pubkey *btcec.PublicKey, network *network.Network) Payment {
 	return Payment{network, pubkey, hash, nil, nil, nil}
 }
 
+// FromPayment creates a Payment struct from a another Payment
+func FromPayment(payment *Payment, network *network.Network) Payment {
```
altafan, line 34: The network should be retrieved from the incoming payment, instead of being provided as parameter.

payment/payment.go
```diff
@@ -30,6 +30,13 @@ func FromPublicKey(pubkey *btcec.PublicKey, network *network.Network) Payment {
...
+	buf := payment.Hash
+	hash := hash160(buf)
```
altafan, line 36: Here we should hash the payment `Script`, instead of the `Hash`

sekulicd: It was done based on comment:
https://github.com/vulpemventures/go-elements/issues/9#issuecomment-628507338
"This technically means taking the hash of the given payment and hashing it again to produce the final hash that will be encoded"
@altafan do you think a comment should be updated form 'taking the hash' to 'taking the script'?
Please review if PR is ok now.

altafan: Yes, thanks for pointing this out. Already updated the comment.

tiero: @altafan When the Script file it's valorized? The FromPublicKey, FromPubKeys etc... should do that then?

PR #29, opened 2020-05-15
Payment script hash

### LINKED ISSUE #11: [Payment] ScriptHash

sekulicd:
Before there was no option to derive P2SH address.

This introduces a method that returns base58 P2SH address from Payment.

It closes #11 

Bonus: This fixes the `Script` created  within `FromPublicKey` that by adding a byte for the length of the pubkey hash.


payment/payment_test.go
```diff
@@ -29,3 +29,12 @@ func TestSegwitAddress(t *testing.T) {
 		t.Errorf("TestSegwitAddress: error when encoding segwit")
 	}
 }
+
+func TestScriptHash(t *testing.T) {
+	_, publicKey := btcec.PrivKeyFromBytes(btcec.S256(), privateKeyBytes)
+	p2wpkh := payment.FromPublicKey(publicKey, &network.Regtest)
+	pay := payment.FromPayment(&p2wpkh)
```
altafan, line 36: When running the tests, if you try to log the `Script` that is hashed within the `FromPayment` you can see that it is an empty byte slice.
Since the `hash160` function accepts even empty byte slices it seems that this is working as expected, but it is not actually.

We should add some changes to the `FromPublicKey` method to make this work correctly.

I just opened a ticket for this, hope you can find it clear.

@sekulicd

altafan:
@sekulicd it would be nice if you could change the return type of the `FromPublicKey` method so it returns a `*Payment` rather than `Payment`.

payment/payment_test.go
sekulicd: Please @altafan can you review this after changes?

payment/payment.go
```diff
@@ -28,19 +28,19 @@ type Payment struct {
 // pub 036f5646ed688b9279369da0a4ad78953ae7e6d300436ca8a3264360efe38236e3
 
 // FromPublicKey creates a Payment struct from a btcec.publicKey
-func FromPublicKey(pubkey *btcec.PublicKey, network *network.Network) Payment {
+func FromPublicKey(pubkey *btcec.PublicKey, network *network.Network) *Payment {
 	publicKeyBytes := pubkey.SerializeCompressed()
 	hash := hash160(publicKeyBytes)[:ripemd160.Size]
 	script := make([]byte, 0)
 	script = append([]byte{Op0}, hash...)
```
altafan, line 35: Just made some tests to cross-check that the segwit addresses produced with go-elements are the same of those returned by the liquidjs-lib.
Of course, the results do not match 😅.
This is because the witness pubkey script misses 1 byte, that is the length of the hash, so this should be changed as follows:
```suggestion
      script = append([]byte{Op0, byte(len(hash))}, hash...)
```


PS. For the test, the expected address is `XZavBojABpfXhPWkw7y9YYFNAezUHZR47m`

altafan: @sekulicd, please, fix this, then we can merge this 👍

sekulicd: Done.

Issue #12, opened 2020-05-12
[Payment] WitnessScriptHash


altafan:
A `WitnessScriptHash` method similar to the `WitnessPubKeyHash` one must be added to the `*Payment` type in order to return a bech32 encoded p2wsh address string, generated from either Hash or from Redeem (in order of least effort).

Ref: #8.

sekulicd:
@altafan please advise if bellow test case corresponds to task requirement:
```
p2wpkh := payment.FromPublicKey(publicKey,Regtest)
redeem := payment.FromPayment(p2wpkh,Regtest)
p2wsh := redeem.WitnessScriptHash()
```

If this is the case we would have two methods which are returning valid values only if FromPayment factory method is called before their invocation:
```
ScriptHash()
WitnessScriptHash()
```
, if this is true, would it be better if those two methods have Payment as an input param? 
, so that we don't allow a user of the package to do below:
```
p2pkh/p2wpkh := payment.FromPublicKey
p2sh/p2wsh := payment.ScriptHash()/payment.WitnessScriptHash()
```
which would return wrong values.

altafan:
Yes,  the example is correct. Thanks for pointing this out. 

Instead of passing the payment as an argument to the `ScriptHash` and `WitnessScriptHash` functions, we force them on working only with "nested payments".
In our implementation, nested payments are those payments that have a non-nil `Redeem` field.

The factory methods `FromPayment` and `FromScript` are intended to return nested payment.
I'm going to add a comment to #10 to better explain how to accomplish this goal.

altafan:
We should definitely add a checkpoint to actually be sure that `Redeem` and `Redeem.Script` are not `nil` into `ScriptHash` and `WitnessScriptHash` methods.  This would prevent unexpected behaviors when calling these functions.

Issue #10, opened 2020-05-12
[Payment] FromScript


altafan:
A `FromScript` function similar to the existing `FromPublicKey` one must be added to the `payment` package in order to return a `Payment` from an output script and a network.

altafan:
Taking into account the considerations pointed in https://github.com/vulpemventures/go-elements/issues/12#issuecomment-629326718, the `FromScript` method actually needs to return a _nested payment_.

This means that firstly the incoming script should be used to create a _redeem_ payment, then the _redeem_ would be passed to the `FromPayment` function.

Example:

```
func FromScript(script []byte) (*Payment, error) {
   redeem := &Payment{Script: script}
   return FromPayment(redeem)
}
```

In this way, we would be  sure that the returned payment has a non `nil` `Redeem` and that the `ScriptHash` and `WitnessScriptHash` function would return the correct addresses.

tiero:
This should NOT be taken before the fix to be done described in #35

PR #36, opened 2020-05-18
Payment-FromPayment does not set a redeem

### LINKED ISSUE #35: [Payment] FromPayment does not set a Redeem
The `FromPayment` function is not actually setting a `Redeem` for the returned payment.
https://github.com/vulpemventures/go-elements/blob/master/payment/payment.go#L43
Instead of `nil`, we should use the `payment` function argument.

sekulicd:
Before the factory method, FromPayment was not setting Redeem field.

After the factory method, FromPayment sets the Redeem field.
Removed test case with a plan to add it as a part of other tasks #10 and #12.

It closes #35 

### DIFF
--- a/payment/payment.go
+++ b/payment/payment.go
@@ -2,7 +2,6 @@ package payment
 
 import (
 	"crypto/sha256"
-	"errors"
 	"github.com/btcsuite/btcd/btcec"
 	"github.com/vulpemventures/go-elements/address"
 	"github.com/vulpemventures/go-elements/network"
@@ -45,12 +44,6 @@ func FromPayment(payment *Payment) *Payment {
 	return &Payment{nil, nil, nil, nil, redeem, nil}
 }
 
-// FromScript creates a Payment struct from a script
-func FromScript(script []byte) (*Payment, error) {
-	redeem := &Payment{Script: script}
-	return FromPayment(redeem), nil
-}
-
 // PubKeyHash is a method of the Payment struct to derive a base58 p2pkh address
 func (p *Payment) PubKeyHash() string {
 	payload := &address.Base58{p.Network.PubKeyHash, p.Hash}
@@ -71,13 +64,10 @@ func (p *Payment) WitnessPubKeyHash() string {
 }
 
 // ScriptHash is a method of the Payment struct to derive a base58 p2sh address
-func (p *Payment) ScriptHash() (string, error) {
-	if p.Redeem == nil {
-		return "", errors.New("redeem not set")
-	}
-	payload := &address.Base58{p.Redeem.Network.ScriptHash, p.Redeem.Hash}
+func (p *Payment) ScriptHash() string {
+	payload := &address.Base58{p.Network.ScriptHash, p.Hash}
 	addr := address.ToBase58(payload)
-	return addr, nil
+	return addr
 }
 
 // Calculate the hash of hasher over buf.
--- a/payment/payment.go
+++ b/payment/payment.go
@@ -2,6 +2,7 @@ package payment
 
 import (
 	"crypto/sha256"
+	"errors"
 	"github.com/btcsuite/btcd/btcec"
 	"github.com/vulpemventures/go-elements/address"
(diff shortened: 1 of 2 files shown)


payment/payment.go
```diff
@@ -40,7 +40,8 @@ func FromPublicKey(pubkey *btcec.PublicKey, network *network.Network) *Payment {
 func FromPayment(payment *Payment) *Payment {
 	buf := payment.Script
 	hash := hash160(buf)
-	return &Payment{payment.Network, payment.PublicKey, hash, nil, nil, nil}
+	redeem := &Payment{payment.Network, payment.PublicKey, hash, nil, nil, nil}
+	return &Payment{nil, nil, nil, nil, redeem, nil}
```
altafan, line 44: I rather return something like:
```
return &Payment{payment.Network, nil, hash, nil, payment, nil}
```

tiero: Also this way you are rewriting the Hash that it's already present in given payment in the redeem field. 

Eg. In case of `payment` having already a `Hash` beacuse it's a `pay to witness pubkey hash` this way you are reqriting it with an extra hash160 that's only neede by ScriptHash construction.

I'm right @altafan it's better preserve the original given payment in the `Redeem` field?

payment/payment.go
```diff
@@ -40,7 +40,8 @@ func FromPublicKey(pubkey *btcec.PublicKey, network *network.Network) *Payment {
 func FromPayment(payment *Payment) *Payment {
 	buf := payment.Script
 	hash := hash160(buf)
-	return &Payment{payment.Network, payment.PublicKey, hash, nil, nil, nil}
+	redeem := &Payment{payment.Network, payment.PublicKey, hash, nil, nil, nil}
```
tiero, line 43:
```suggestion
	redeem := &Payment{payment.Network, payment.PublicKey, payment.Hash, nil, nil, nil}
```

tiero: Is this correct @altafan?

payment/payment.go
altafan: Yes, the incoming `payment` doesn't and shouldn't be modified but just used to produce the hash of the `*Payment` to return, as it's already done, and also included in it in the `Redeem` field.

PS: I think my example is not the best option because by using the same `payment` pointer  for the return value of the `FromPayment` method, this could lead to unexpected behaviors if for any reason the `payment` pointer is deleted. I think it would be better to make a copy and use a pointer to a different memory allocation.

payment/payment.go
```diff
@@ -38,10 +38,8 @@ func FromPublicKey(pubkey *btcec.PublicKey, network *network.Network) *Payment {
 
 // FromPayment creates a Payment struct from a another Payment
 func FromPayment(payment *Payment) *Payment {
-	buf := payment.Script
-	hash := hash160(buf)
-	redeem := &Payment{payment.Network, payment.PublicKey, hash, nil, nil, nil}
-	return &Payment{nil, nil, nil, nil, redeem, nil}
+	redeem := &Payment{payment.Network, payment.PublicKey, payment.Hash, nil, nil, nil}
```
altafan, line 41: `redeem` looks better now, but I would copy all the fields (even if `nil`) rahter than forcing them to `nil`:
```suggestion
	redeem := &Payment{payment.Network, payment.PublicKey, payment.Hash, payment.BlindingKey, payment.Redeem, payment.Script}
```

payment/payment.go
```diff
@@ -38,10 +38,8 @@ func FromPublicKey(pubkey *btcec.PublicKey, network *network.Network) *Payment {
...
+	return &Payment{payment.Network, nil, payment.Hash, nil, redeem, nil}
```
altafan, line 42: The `Hash` of the payment should be, as in the previous commit, the hash of the redeem script:
```suggestion
	return &Payment{payment.Network, nil, hash160(redeem.Script), nil, redeem, nil}
```

altafan: Because of this, even before instantiating `redeem` we should check that `payment.Script != nil`, returning an error otherwise.

PR #37, opened 2020-05-18
Pset: creator, updater, signer, finalizer and extractor

### LINKED ISSUE #27: [Pset] Creator
The role of the creator is to craft a partial elements transaction starting from:
* version number
* list of inputs 
* list of outputs
* locktime

reference implementation: https://github.com/btcsuite/btcutil/blob/master/psbt/creator.go

### LINKED ISSUE #28: [Pset] Updater
The role of the updater is to update the fields of the partial inputs and outputs of a partial transaction.

The `Updater`, thus, has a unique field of type `*Pset`, and exposes methods `AddIn*` `AddOut*` as setters for every field of `PInput` and `POutput`.

Reference implementation: https://github.com/btcsuite/btcutil/blob/master/psbt/updater.go

### LINKED ISSUE #31: [Pset] Signer
The updater is also in charge of providing a method `Sign` to sign input of a partial elements transaction, but we keep this function in a separated file `signer.go`.

Reference implementation: https://github.com/btcsuite/btcutil/blob/master/psbt/signer.go.

This needs to be reimplemented since the reference implementation expects to work with `wire.TxIn` that slighlty differs from our `transaction.TxInput`.

### LINKED ISSUE #32: [Pset] Finalizer
The role of the finalizer is to remove all temporary data from partial inputs when they've all been signed correctly.

Reference implementation: https://github.com/btcsuite/btcutil/blob/master/psbt/finalizer.go.

We cannot import the reference implementation since when finalizing, it creates new partial inputs with just the witness and signature fields. We would end up having Bitcoin version of partial inputs in this way.
We do not want of course this to happen so we need to re-write the `finalizer.go` using our partial input methods.

### LINKED ISSUE #33: [Pset] Extractor
The role of the extractor is to retrieve the serialized final signed transaction from a partial transaction. Even if the `Extract` function actually will be an exported method of `*Pset`, we should keep this in a dedicated `extractor.go` file maintain separation of roles.

Reference implementation: https://github.com/btcsuite/btcutil/blob/master/psbt/extractor.go.

The main difference from the reference implementation is that the method will return a `*transaction.Transaction` pointer rather then a `*wire.MsgTx` one, thus using the elements transaction serialization instead of the bitcoin one.

altafan:
Before this, the `pset` package was helpful only for serializing and deserializing a partial elements transaction.

This adds the features for creating, updating, signing, and finalizing a partial transaction. Last but not least, this introduces a feature for extracting the final signed transaction from a pset too.

### Creator

The creator is an exported factory function named  just `New` that accepts a list of tx inputs, a list of tx outputs, a tx version and locktime, and returns a new instance of `*Pset` containing an unsigned transaction.

### Updater

The updater, as the name suggests, has the responsibility of updating the fields of any partial input or output. Thus, basically, for any Field of the `PInput` and `POutput` type, the `Updater` type exposes a function `AddIn<PInputfieldName>` and `AddOut<POutputFieldName>`.
To create a new updater, a factory function `NewUpdater` accepting a `*Pset` instance and returning a `*Updater` one is exposed.

### Signer

The signer is in charge of checking that when adding a signature to an input of the pset, this is valid and that also the pset is correctly structured.
Given that, this role is implemented as a function `Sign` of the `*Updater` type.
This function accepts an input index, a signature, a public key, and one between a redeem or witness script and checks that the signature is valid against the given script and pubkey, along with setting the partial input's script to the one provided.

### Finalizer

The finalizer takes a partial transaction and combines every input `PartialSignature` into the final input's `SignatureScript`. After finalizing, the partial transaction is _complete_, and it's ready to be extracted from the Pset wrapper and broadcasted to the network.

This role is accomplished by a `Finalize` function that accepts a `*Pset` instance and an input index and performs the process described above, returning an error if any occurs during the process. It previously checks that the provided pset is valid in the sense that it's ready to be finalized; otherwise, an error is returned.

A handy `FnalizeAll` that runs the above method for every input of the provided `*Pset` is also exported.

### Extractor

The extractor is a simple `Extract` function that accepting a `*Pset` instance that checks that the provided partial transaction is complete (it returns an error otherwise), then make a copy of the pset's `UnsignedTx` and for every partial input adds the final input `SignatureScript` and `PublicKey` to the input witnesses. After this, the `*Transaction` instance is returned.

This closes #27.
This closes #28.
This closes #31.
This closes #32.
This closes #33.

Please @tiero, review this.


pset/pset_test.go
```diff
@@ -63,3 +67,235 @@ func testRoundTrip(t *testing.T) {
...
+}
+
+func toConfidentialValue(val int) ([]byte, error) {
+	unconfPrefix := byte(1)
+	b := bytes.NewBuffer([]byte{})
+	if err := transaction.BinarySerializer.PutUint64(b, binary.LittleEndian, uint64(val)); err != nil {
+		return nil, err
+	}
+	return append([]byte{unconfPrefix}, reverseBytes(b.Bytes())...), nil
+}
```
tiero, line 301: Could you add an end to end test also? @altafan

tiero:
Could you copy the description of this PR in the `pset/README.md` ?

pset/utils.go
```diff
@@ -371,3 +371,19 @@ func validateSignature(sig []byte) bool {
 func checkValid(ps psbt.PartialSig) bool {
 	return validatePubkey(ps.PubKey) && validateSignature(ps.Signature)
 }
+
+// reverseBytes returns the given byte slice with elems in reverse order.
+func reverseBytes(buf []byte) []byte {
```
tiero, line 376: Why not `sort.Reverse` method?

altafan: The functions you mentioned work with signatures,  to eventually change the order of them. The goal of `reverseBytes` function, instead, is to reverse a byte slice. This is necessary when encoding the transaction hash. I don't understand the correlation between this helper and the ones of the partial signature sorter.

pset/pset_test.go
altafan: What do you mean by "adding an end"?

tiero: e2e test. From creation to extraction in a single one-shot function?

tiero:
@altafan  Minor: add the ticket being closed in the description.

pset/pset_test.go
tiero: Also I noticed more failing test cases in the `psbt` repo. We'll add more coverage later?

PR #38, opened 2020-05-18
FromScript factory method

### LINKED ISSUE #10: [Payment] FromScript

sekulicd:
Before the factory method, FromScript has added it was not possible to create nested Payment based on script.

After the factory method, FromScript returns nested Payment based on script.

It closes #10


payment/payment.go
```diff
@@ -46,6 +46,12 @@ func FromPayment(payment *Payment) (*Payment, error) {
 	return &Payment{payment.Network, nil, hash160(redeem.Script), nil, redeem, nil}, nil
 }
 
+// FromPayment creates a nested Payment struct from script
+func FromScript(script []byte) (*Payment, error) {
```
altafan, line 50: I would add a `network` argument. If its value is `nil`, it should default to the liquid mainnet network.

altafan: This in order to not have a nil `Network` in the payment. This field is important because the prefix of an address changes based on the network.

tiero: We would check this for consistency also in the FromPublicKey factory function too then?

sekulicd: I added validation in both methods and renamed input param name from 'network' to 'net',
I think we should avoid naming variables same as imported package name, do you agree with this?

payment/payment.go
```diff
@@ -47,8 +53,14 @@ func FromPayment(payment *Payment) (*Payment, error) {
 }
 
 // FromPayment creates a nested Payment struct from script
-func FromScript(script []byte) (*Payment, error) {
-	redeem := &Payment{Script: script}
+func FromScript(net *network.Network, script []byte) (*Payment, error) {
```
altafan, line 56: just for consistency, change the order of the arguments so that it's the same of the `FromPublicKey` function.
```suggestion
func FromScript(script []byte, net *network.Network) (*Payment, error) {
```

sekulicd: Done.

altafan:
@sekulicd can you add the missing commits from master? Although there are no conflicts, your branch is 3 commits behind vulpemventures/master

PR #40, opened 2020-05-19
p2wsh address

### LINKED ISSUE #12: [Payment] WitnessScriptHash

sekulicd:
Before there was no option to get p2wsh address.

After the method WitnessScriptHash added user of a package can get payment's p2wsh address.
FromPublicKey factory method is updated so that it contains WitnessHash.

It closes #12 

### DIFF
--- a/payment/payment.go
+++ b/payment/payment.go
@@ -22,6 +22,7 @@ type Payment struct {
 	BlindingKey *btcec.PublicKey
 	Redeem      *Payment
 	Script      []byte
+	WitnessHash []byte
 }
 
 //target mail duty exit light void budget zone senior tag rude wisdom
@@ -37,10 +38,11 @@ func FromPublicKey(pubkey *btcec.PublicKey, net *network.Network) *Payment {
 		tmpNet = net
 	}
 	publicKeyBytes := pubkey.SerializeCompressed()
-	hash := hash160(publicKeyBytes)[:ripemd160.Size]
+	pkHash := hash160(publicKeyBytes)[:ripemd160.Size]
 	script := make([]byte, 0)
-	script = append([]byte{Op0, byte(len(hash))}, hash...)
-	return &Payment{tmpNet, pubkey, hash, nil, nil, script}
+	script = append([]byte{Op0, byte(len(pkHash))}, pkHash...)
+	witnessHash := sha256.Sum256(script)
+	return &Payment{tmpNet, pubkey, pkHash, nil, nil, script, witnessHash[:]}
 }
 
 // FromPayment creates a Payment struct from a another Payment
@@ -48,8 +50,8 @@ func FromPayment(payment *Payment) (*Payment, error) {
 	if payment.Script == nil {
 		return nil, errors.New("script can't be nil")
 	}
-	redeem := &Payment{payment.Network, payment.PublicKey, payment.Hash, payment.BlindingKey, payment.Redeem, payment.Script}
-	return &Payment{payment.Network, nil, hash160(redeem.Script), nil, redeem, nil}, nil
+	redeem := &Payment{payment.Network, payment.PublicKey, payment.Hash, payment.BlindingKey, payment.Redeem, payment.Script, payment.WitnessHash}
+	return &Payment{payment.Network, nil, hash160(redeem.Script), nil, redeem, nil, nil}, nil
 }
 
 // FromPayment creates a nested Payment struct from script
@@ -71,6 +73,13 @@ func (p *Payment) PubKeyHash() string {
 	return addr
 }
 
+// ScriptHash is a method of the Payment struct to derive a base58 p2sh address
+func (p *Payment) ScriptHash() string {
+	payload := &address.Base58{p.Network.ScriptHash, p.Hash}
+	addr := address.ToBase58(payload)
+	return addr
+}
+
 // WitnessPubKeyHash is a method of the Payment struct to derive a base58 p2wpkh address
 func (p *Payment) WitnessPubKeyHash() string {
 	//Here the Version for wpkh is always 0
@@ -83,10 +92,15 @@ func (p *Payment) WitnessPubKeyHash() string {
 	return addr
 }
 
-// ScriptHash is a method of the Payment struct to derive a base58 p2sh address
-func (p *Payment) ScriptHash() string {
-	payload := &address.Base58{p.Network.ScriptHash, p.Hash}
-	addr := address.ToBase58(payload)
+// WitnessScriptHash is a method of the Payment struct to derive a base58 p2wsh address
+func (p *Payment) WitnessScriptHash() string {
+	//Here the Version for wpkh is always 0
+	version := byte(0x00)
+	payload := &address.Bech32{p.Network.Bech32, version, p.WitnessHash}
+	addr, err := address.ToBech32(payload)
+	if err != nil {
+		return ""
+	}
 	return addr
 }
 
--- a/payment/payment.go
+++ b/payment/payment.go
@@ -47,15 +47,19 @@ func FromPublicKey(pubkey *btcec.PublicKey, net *network.Network) *Payment {
 
 // FromPayment creates a Payment struct from a another Payment
 func FromPayment(payment *Payment) (*Payment, error) {
-	if payment.Script == nil {
-		return nil, errors.New("script can't be nil")
+	if payment.Script == nil || len(payment.Script) == 0 {
+		return nil, errors.New("payment's script can't be empty or nil")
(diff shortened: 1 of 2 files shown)


payment/payment.go
```diff
@@ -37,19 +38,20 @@ func FromPublicKey(pubkey *btcec.PublicKey, net *network.Network) *Payment {
...
 
 // FromPayment creates a Payment struct from a another Payment
 func FromPayment(payment *Payment) (*Payment, error) {
 	if payment.Script == nil {
 		return nil, errors.New("script can't be nil")
 	}
-	redeem := &Payment{payment.Network, payment.PublicKey, payment.Hash, payment.BlindingKey, payment.Redeem, payment.Script}
-	return &Payment{payment.Network, nil, hash160(redeem.Script), nil, redeem, nil}, nil
+	redeem := &Payment{payment.Network, payment.PublicKey, payment.Hash, payment.BlindingKey, payment.Redeem, payment.Script, payment.WitnessHash}
+	return &Payment{payment.Network, nil, hash160(redeem.Script), nil, redeem, nil, nil}, nil
```
altafan, line 54: The `WitnessHash` should not be nil but the sha256 hash of `redeem.Script`:
```suggestion
	return &Payment{payment.Network, nil, hash160(redeem.Script), nil, redeem, nil, sha256.Sum256(redeem.Script)}, nil
```

payment/payment.go
```diff
@@ -37,19 +38,20 @@ func FromPublicKey(pubkey *btcec.PublicKey, net *network.Network) *Payment {
...
 	}
 	publicKeyBytes := pubkey.SerializeCompressed()
-	hash := hash160(publicKeyBytes)[:ripemd160.Size]
+	pkHash := hash160(publicKeyBytes)[:ripemd160.Size]
 	script := make([]byte, 0)
-	script = append([]byte{Op0, byte(len(hash))}, hash...)
-	return &Payment{tmpNet, pubkey, hash, nil, nil, script}
+	script = append([]byte{Op0, byte(len(pkHash))}, pkHash...)
+	witnessHash := sha256.Sum256(script)
+	return &Payment{tmpNet, pubkey, pkHash, nil, nil, script, witnessHash[:]}
```
tiero, line 45: If I recall correctly the witness program must be 20bytes long, so would be better to bound the size of the hash

tiero: @altafan

sekulicd: witnessProgram is 20bytes for P2WPKH and 32bytes for P2WSH

payment/payment.go
```diff
@@ -83,10 +92,15 @@ func (p *Payment) WitnessPubKeyHash() string {
...
 }
 
-// ScriptHash is a method of the Payment struct to derive a base58 p2sh address
-func (p *Payment) ScriptHash() string {
-	payload := &address.Base58{p.Network.ScriptHash, p.Hash}
-	addr := address.ToBase58(payload)
+// WitnessScriptHash is a method of the Payment struct to derive a base58 p2wsh address
+func (p *Payment) WitnessScriptHash() string {
+	//Here the Version for wpkh is always 0
+	version := byte(0x00)
```
altafan, line 98: Instead of hardcoding the version, you can get it from the very first byte of `p.Script` since we already hardcode it in there:
```suggestion
	version := p.Script[0]
```

Also, in all these functions returning an address, I guess it would be better to add a check for being sure that `Hash`/`WitnessHash`, and also `Script` for this specific case, are not `nil` or empy byte slices. What do you think?

altafan (changes requested):

altafan:
Please, also add the `TestP2SH` you removed with the previous PR.

altafan:
Even if the `TestP2WSH` returns an address, the test itself has no actual mean.

When creating a payment from a public key, a user would then call the `PublicKeyHash()` or `WitnessPublicKeyHash()` at his will. It could even wrap this payment into another one to then call `ScriptHash()` to get the p2sh-p2wpkh address.

But the address in the form `p2wsh-p2pkh`, as done in `TestP2WSH`, is not a usable address and could cause loss of funds if used!

To properly test a p2wsh address, you can hardcode this multisig redeem script (here hex encoded):
```
52410479be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798483ada7726a3c4655da4fbfc0e1108a8fd17b448a68554199c47d08ffb10d4b84104c6047f9441ed7d6d3045406e95c07cd85c778e4b8cef3ca7abac09b95c709ee51ae168fea63dc339a3c58419466ceaeef7f632653266d0e1236431a950cfe52a52ae
```
and use it to create a p2wsh, exploiting the `FromScript` factory function:
```
p2ms, err := FromScript(redeemScript, &network.Regtest)
p2wsh, err := FromPayment(p2ms)
address := p2wsh.WitnessScriptHash()
```

The `address` should be equal to:
```
ert1q2z45rh444qmeand48lq0wp3jatxs2nzh492ds9s5yscv2pplxwesajz7q3
```

@sekulicd

sekulicd:
@altafan, regarding comment [here](https://github.com/vulpemventures/go-elements/pull/40#issuecomment-630918001), factory method FromPublicKey is creating p2wpkh script [here](https://github.com/vulpemventures/go-elements/blob/a9f4833b091037bdc37efdf1b6baeebb7e6131a9/payment/payment.go#L42) and then this script is hashed and kept in witnessHash field [here](https://github.com/vulpemventures/go-elements/blob/28117980fca504b924ebe9a432e4aaf237a0e329/payment/payment.go#L44) , so in TestP2WSH redeem script is p2wpkh stored in witnessHash and WitnessScriptHash is creating Bech32 address based on WitnessHash, am I right?

altafan:
What I am trying to say is that the type of payment that you are describing is not a valid one.

When you generate an address from your public key, you can create 3 types of addresses: `p2pkh`, `p2sh-p2wpkh`, or `p2wpkh`. But not a `p2wsh-p2pkh` or `p2wsh-p2wpkh` as done in the test.

I would say that all the methods of the `payment` package work in the correct way (after requested changes), but in the `TestP2WSH` test, at the moment, we're just not using them in the right manner.

One good test case, instead, should be the example I made in my previous comment, where we generate the p2wsh address from multisig script. 

If you don't know multisigs, this type of scripts are composed of `N` multiple public keys instead of just one, letting people create wallets with more participants where a minimum `M` number of them need to sign a transaction to correctly spend funds.
So you may hear about `M-of-N` multisig wallet/scripts.

The one I provided in the previous comment is the script of a `2-of-2` wallet. That is the hex encoding of this bitcoin script:
```
OP_2 // this is M
0479be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798483ada7726a3c4655da4fbfc0e1108a8fd17b448a68554199c47d08ffb10d4b8 // pubkey 1
04c6047f9441ed7d6d3045406e95c07cd85c778e4b8cef3ca7abac09b95c709ee51ae168fea63dc339a3c58419466ceaeef7f632653266d0e1236431a950cfe52a // pubkey 2
OP_2 // this is N
OP_CHECKMULTISIG
```.
This means that the script is formed by 2 public keys and both of the respective private keys need to sign a transaction when spending the funds locked by this script.

Regarding p2wsh addresses, this is the most common one used in the bitcoin network.

@sekulicd

sekulicd:
@altafan, I understand how multising works, 
you have 'M pk pk pk ... N CHECKMULTISIG' script which is then hashed and used to create p2sh/p2wsh.
What I also understood is that in similar way addresses like p2sh or p2wsh can be created from the hash of any type of redeeming script (even one that is not used in bitcoin network eg, '3 3 EQUAL').
Since in FromPublicKey we have bellow two lines:
`script = append([]byte{Op0, byte(len(pkHash))},pkHash...) -> p2wpkh`
`witnessHash := sha256.Sum256(script)`
Script field holds p2wpkh script and you are right address created in this test case would be something like p2wpkh - p2wsh.
I just thought that you can pass any redeem script and generate p2sh or p2wsh
Anyway tnx for the explanation, I will adjust the test cases in the morning.

payment/payment.go
sekulicd: @altafan this is only the case if we create a payment using factory method FromPublicKey.

altafan: Yes, you're right 😄. Ignore this change.

altafan (approved):

PR #46, opened 2020-05-20
added factory method FromPublicKeys for creating multiscript payment

### LINKED ISSUE #43: [Payment] FromPublicKeys factory
The `payment` package is almost complete. I would say that it still misses a factory function for creating a payment given multiple public keys in order to create a payment for a multisig script.

Given that, we should add a `FromPublicKeys` method that accepts a list of public keys `pubkeys []*btcec.PublicKey`, a threshold `m int`  and of course a network `net *network.Network`. The first 2 arguments should be used to create the multisig script:
```
OP_<M> pubkeys OP_<N> OP_CHECKMULTISIG
```

and then use the `FromScript` factory to return a new `*Payment`.

sekulicd:
Before the factory method, FromPublicKeys has added it was not possible to create multi-signature Payment.

After the factory method, FromPublicKeys returns nested Payment with multi-signature script value inside Script field.

It closes #43 

### DIFF
--- a/payment/payment.go
+++ b/payment/payment.go
@@ -3,7 +3,9 @@ package payment
 import (
 	"crypto/sha256"
 	"errors"
+	"fmt"
 	"github.com/btcsuite/btcd/btcec"
+	"github.com/btcsuite/btcd/txscript"
 	"github.com/vulpemventures/go-elements/address"
 	"github.com/vulpemventures/go-elements/network"
 	"golang.org/x/crypto/ripemd160"
@@ -11,7 +13,8 @@ import (
 )
 
 const (
-	Op0 = 0x00
+	Op0             = 0x00 // 0
+	OpCheckMultiSig = 0xae // 174
 )
 
 // Payment defines the structure that holds the information different addresses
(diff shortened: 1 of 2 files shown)


payment/payment.go
```diff
@@ -3,15 +3,18 @@ package payment
...
 	"github.com/vulpemventures/go-elements/network"
 	"golang.org/x/crypto/ripemd160"
 	"hash"
 )
 
 const (
-	Op0 = 0x00
+	Op0             = 0x00 // 0
+	OpCheckMultiSig = 0xae // 174
 )
```
altafan, line 18: We should use the opcodes from btcd instead of redefining them into payment https://github.com/btcsuite/btcd/blob/86fed781132ac890ee03e906e4ecd5d6fa180c64/txscript/opcode.go

payment/payment.go
```diff
@@ -45,6 +48,37 @@ func FromPublicKey(pubkey *btcec.PublicKey, net *network.Network) *Payment {
 	return &Payment{tmpNet, pubkey, pkHash, nil, nil, script, witnessHash[:]}
 }
 
+// FromPublicKeys creates a multi-signature Payment struct from list of public key's
+func FromPublicKeys(pubkeys []*btcec.PublicKey, numOfPksRequired int, net *network.Network) (*Payment, error) {
```
altafan, line 52: I would call the threshold `m` or `nrequired` as done in btcutil.

payment/payment_test.go
```diff
@@ -72,3 +76,18 @@ func TestP2WSH(t *testing.T) {
...
+	_, publicKey1 := btcec.PrivKeyFromBytes(btcec.S256(), privateKeyBytes1)
+	_, publicKey2 := btcec.PrivKeyFromBytes(btcec.S256(), privateKeyBytes2)
+
+	p2ms, err := payment.FromPublicKeys([]*btcec.PublicKey{publicKey1, publicKey2}, 1, &network.Regtest)
+	if err != nil {
+		t.Error(err)
+	}
+
+	if hex.EncodeToString(p2ms.Script) != "5121036f5646ed688b9279369da0a4ad78953ae7e6d300436ca8a3264360efe38236e321023c61f59e9a3a3eb01c3ed0cf967ad217153944bcf2498a8fd6e70b27c7ab6ee652ae" {
+		t.Error("hax value of p2ms script not as expected")
```
altafan, line 90: Typo
```suggestion
		t.Error("hex value of p2ms script not as expected")
```

I would also test both the ScriptHash and WitnessScriptHash addresses for this payment

altafan: the resulting addresses should be:
* ScriptHash: `XJkohBHRMT8JUknSqCH7aJP9gAuAe9eNLY`
* WitnessScriptHash: `ert1q484pt3gqgthcxa35nl4t6utpd0uf7tkm240hlxe6k4newkydwcqs5sjc4c`

PR #50, opened 2020-05-21
Transaction enhancements & Pset integration test 

### LINKED ISSUE #39: Let a user spend an unconfidential input and broadcast the transaction

### LINKED ISSUE #47: [Transaction] HashForSignature method
At the moment, the `transaction` package provides methods for just serializing a transaction.
It is not possible to get a serialization based on a sighash type in order to generate an hash used to produce a signature for an input.

In _btcd_ this is accomplished by the `calcSignatureHash` [function](https://github.com/btcsuite/btcd/blob/86fed781132ac890ee03e906e4ecd5d6fa180c64/txscript/script.go#L599) of the _txscript_ package.

Since this implementation works with `msgTx` type (bitcoin transaction), we need to reimplement it, and I would suggest doing so by adding an exported `HashForSignature` method to the `*Transaction` type.
Something similar is already done in [liquidjs-lib](https://github.com/vulpemventures/liquidjs-lib/blob/master/ts_src/transaction.ts#L380).

### LINKED ISSUE #48: [Transaction] HashForWitnessV0 mehtod
Taking into account the considerations of #47, we should also add a `HashForWitnessV0` method that serializes and hashes the transaction so that it would be then possible to use it for producing a witness signature, according to the [BIP-0143](https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki#specification) specification.

altafan:
This adds an intergration test for the _pset_ package.
It is worth to note that this is an example of a **unconfidential** transaction.

The `TestFromCreateToBroadcast` ensures that by using the methods of the _pset_  package, one is able to create a transaction from scratch, to update it in order to have a complete signed transaction and, lastly, to broadcast it to the network so that it's included in the Elements blockchain.

Since the signer role of the _pset_ package does not actually handle the signing process, but rather is responsible of updating and checking the validity of the signature scripts of the partial inputs, this also adds 2 new exported methods to the `*Transaction` type of the _transaction_ package in order to let a user correctly sign transaction inputs.

Before this, there was no way to serialize a transaction in order to hash it and use this hash for producing a signature (with also a private key and a pubkey script of course).
This adds the methods `HashForSignature`, which returns  the hash of the legacy tx serialization for signing an input, and `HashForWitnessV0`, which, instead, returns the hash of the segwit tx serialization for signing an input, according to the BIP-0143 specification (some changes compared to the spec were necessary since Elements inputs and outputs slightly differ from the Bitcoin ones).

NOTE:
Now, when running the tests it is mandatory to export an `API_URL` environment variable pointing to the nigiri-chopsitcks url. For development, it is mandatory to have a running local Nigiri instance or at least a remote one accessible from the internet.
Travis yaml has been updated to download and use the travis-nigiri tool so the integration test runs against a local Nigiri instance.

This closes #47.
This closes #48.
This closes #39.

Please @tiero, review this.


altafan:
@tiero, at the moment we have the same `reverseBuffer` util function in 2 different places:
* pset/utils.go
* transaction/common.go
I would change this method to the exported version `ReverseBuffer`, but how should we call the package where this function resides? I would not place this into _pset_ or _transaction_ since we risk to then create circular dependencies, rather we should create a new _util_ package or whatever name. Do you have any suggestion on the name to choose?

altafan:
The `toConfidentialValue` function present both into [`pset_test.go`](https://github.com/vulpemventures/go-elements/blob/8a37f5a5f5affc4f182ae5a9d80911c0efd144c2/pset/pset_test.go#L429) and [`transaction_test.go`](https://github.com/vulpemventures/go-elements/blob/8a37f5a5f5affc4f182ae5a9d80911c0efd144c2/transaction/transaction_test.go#L212) will be moved in the future into a _confidential_ or what-so-ever package when we'll support confidential transactions, so, by now, we should not worry about having this repeated in 2 different test files.

tiero:
> `reverseBuffer`

The best way would be to put in the `internal` folder that's enforced by the compiler from go 1.14 https://github.com/golang-standards/project-layout/tree/master/internal

You could then create a sub-folder containing a package called  simply `reversebuffer`. Toptally fine to have a pakcage for 20 LOC.

.travis.yml
```diff
@@ -1,10 +1,30 @@
 language: go
 
-sudo: false
-
 go:
-  - tip
+  - 1.13.x
```
tiero, line 4: Let's use the latest one? 1.14.x

I personally use 1.14.2 on my machine

.travis.yml
```diff
@@ -1,10 +1,30 @@
...
+
+services:
+  - docker
+
+before_install:
+  # Get latest docker version
```
tiero, line 10: Do we really need the latest Docker version? I recall we needed this for having experimental features like buildx

.travis.yml
```diff
@@ -1,10 +1,30 @@
...
+  - docker-compose -f tmp/docker-compose.yml up -d
+
+env:
+  - API_URL=http://localhost:3001
 
 script:
   - if [ -n "$(gofmt -l .)" ]; then echo "Go code is not formatted"; exit 1; fi
-  - go test -v ./...
\ No newline at end of file
+  - go test -count=1 ./... -v
```
tiero, line 27: could you add -race option?

transaction/transaction.go
```diff
@@ -447,6 +464,145 @@ func (tx *Transaction) Serialize() ([]byte, error) {
...
+	hashSequences := Zero
+	hashIssuances := Zero
+	hashOutputs := Zero
+
+	// Inputs
+	if (hashType & txscript.SigHashAnyOneCanPay) == 0 {
+		hashInputs = calcTxInputsHash(tx.Inputs)
+	}
+	// Sequences
+	if (hashType&txscript.SigHashAnyOneCanPay) == 0 &&
```
tiero, line 562: whitespace?

transaction/transaction.go
```diff
@@ -447,6 +464,145 @@ func (tx *Transaction) Serialize() ([]byte, error) {
...
+		(hashType&0x1f) != txscript.SigHashSingle &&
```
tiero, line 563: whitespaces?

transaction/transaction.go
```diff
@@ -447,6 +464,145 @@ func (tx *Transaction) Serialize() ([]byte, error) {
...
+		(hashType&0x1f) != txscript.SigHashNone {
+		hashSequences = calcTxSequencesHash(tx.Inputs)
+	}
+	// Issuances
+	if (hashType & txscript.SigHashAnyOneCanPay) == 0 {
+		hashIssuances = calcTxIssuancesHash(tx.Inputs)
+	}
+	// Outputs
+	if (hashType&0x1f) != txscript.SigHashSingle &&
```
tiero, line 572: whitespace?

PR #70, opened 2020-06-01
Confidential final value blinding factor

### LINKED ISSUE #62: [Confidential] Final value blinding factor
To better explain what *confidential transaction* means, take a look at [this](https://blockstream.info/liquid/block/2876c0eb52df8568b0e6e48796d29685067ee81fc45dcdc8eae0fcfc6861126f) confidential transaction compared to [this](https://blockstream.info/liquid/tx/511840c9e97b85bdf42b0c96e4f252699ee75c51f4c66586b5ee9d80a5d197ea) unconfidential one.
You can see that in a confidential tx all outputs amount and asset are "hidden" by commitments (except for the network fee out).
Along with these commitments, the confidential utxos, compared to the uncofidential ones, have non-null `nonce`, `rangeProof` and `surjectionProof`, that, from an high-level perspective, are necessary to verify the validity of the commitments and for unblinding the output of course.

The *commitments* cited above, basically, are created starting from the real data to blind (asset/value) and a blinding factor.
For all outputs (excluded the fee one that does must not to be blinded), the asset blinding factors are generated randomly, while the ones for the values are generated randomly, except for the last one that is created with a `FinalValueBlindingFactor` method that calls the binding for the native `secp256k1_pedersen_blind_generator_blind_sum` function.

This method returns a value blinding factor starting from all input and output values and asset, and their value blinding factors.
[Here](https://github.com/vulpemventures/liquidjs-lib/blob/97673986c4233dee61f98dc59b5256ed2a8f313d/ts_src/confidential.ts#L10) you can find a reference implementation of the `FinalValueBlindingFactor` that we need to expose in our *confidential* package, while [here](https://github.com/vulpemventures/liquidjs-lib/blob/97673986c4233dee61f98dc59b5256ed2a8f313d/ts_src/psbt.ts#L736) you can see how the js version of this method is used by the liquidjs-lib pset's blinder role.

sekulicd:
Before there was no option to calculate blind sum.

After the method FinalValueBlindingFactor is added, the user can calculate final value blinding sum.

It closes #62 


confidential/confidential_test.go
```diff
@@ -79,3 +105,92 @@ func TestUnblindOutput(t *testing.T) {
...
+	}
+
+	vectors := tests["FinalValueBlindingFactorTC"].([]interface{})
+	for _, testVector := range vectors {
+		v := testVector.(map[string]interface{})
+
+		inValuesSlice := v["inValues"].([]interface{})
+		inValues := make([]uint64, 0)
+		for _, v := range inValuesSlice {
+			n, err := strconv.ParseUint(fmt.Sprintf("%v", v), 10, 64)
```
altafan, line 122: why not `v.(string)` instead of `fmt.Sprintf()`?

confidential/confidential_test.go
```diff
@@ -79,3 +105,92 @@ func TestUnblindOutput(t *testing.T) {
...
+			if !assert.NoError(t, err) {
+				t.FailNow()
+			}
+			inValues = append(inValues, n)
+		}
+
+		outValuesSlice := v["outValues"].([]interface{})
+		outValues := make([]uint64, 0)
+		for _, v := range outValuesSlice {
+			n, err := strconv.ParseUint(fmt.Sprintf("%v", v), 10, 64)
```
altafan, line 132: same here and in the all the following

confidential/confidential_test.go
```diff
@@ -79,3 +105,92 @@ func TestUnblindOutput(t *testing.T) {
...
+		t.FailNow()
+	}
+
+	vectors := tests["FinalValueBlindingFactorTC"].([]interface{})
+	for _, testVector := range vectors {
+		v := testVector.(map[string]interface{})
+
+		inValuesSlice := v["inValues"].([]interface{})
+		inValues := make([]uint64, 0)
+		for _, v := range inValuesSlice {
```
altafan, line 121: I would change the name of the variable since `v` is already declared before. `vv` is just fine to me.

confidential/data/confidential.json
```diff
@@ -1,5 +1,5 @@
 {
-  "unblind": [
+  "UnblindOutputTC": [
```
altafan, line 2: Use `unblindOutput`. Generally, we use the name of the method to test, but not capitalized.

confidential/data/confidential.json
```diff
@@ -14,5 +14,27 @@
         "assetBlindingFactor": "5dd4b83ac4dc828b033874f818bd0b5575371cc0e1278c554905dfbb49740fda"
       }
     }
+  ],
+  "FinalValueBlindingFactorTC": [
```
altafan, line 18:
```suggestion
  "finalValueBlindingFactor": [
```

confidential/confidential_test.go
sekulicd: Done

confidential/confidential_test.go
sekulicd: I changed it to val.

confidential/data/confidential.json
sekulicd: Done

confidential/data/confidential.json
sekulicd: Done

confidential/confidential_test.go
sekulicd: I changed it to val.

PR #71, opened 2020-06-01
Confidential asset and value commitments

### LINKED ISSUE #63: [Confidential] Value commitment
We need to expose a `ValueCommitment` method that expects a value (expressed in satoshi as `uint64`), an asset commitment and a blinding factor (respectively as `[33]byte` and `[32]byte`) and returns the `[33]byte` value commitment.
This method basically calls the binding for the native `secp256k1_pedersen_commit` function.

[This](https://github.com/vulpemventures/liquidjs-lib/blob/97673986c4233dee61f98dc59b5256ed2a8f313d/ts_src/confidential.ts#L30) can be used as reference implementation, while [here](https://github.com/vulpemventures/liquidjs-lib/blob/97673986c4233dee61f98dc59b5256ed2a8f313d/ts_src/psbt.ts#L770) you can see how this is used by the blinder role of the liquidjs-lib's pset.

This actually depends on #64.

### LINKED ISSUE #64: [Confidential] Asset commitment
We need to expose a `AssetCommitment` method that expects an asset id (as `[]byte`) and a blinding factor (as `[32]byte`) and returns the [33]byte asset commitment.
This method basically calls the binding for the native `secp256k1_generator_generate_blinded` function.

[This](https://github.com/vulpemventures/liquidjs-lib/blob/97673986c4233dee61f98dc59b5256ed2a8f313d/ts_src/confidential.ts#L40) can be used as reference implementation, while [here](https://github.com/vulpemventures/liquidjs-lib/blob/97673986c4233dee61f98dc59b5256ed2a8f313d/ts_src/psbt.ts#L766) you can see how this is used by the blinder role of the liquidjs-lib's pset.

sekulicd:
Before there was no option to create commitment for asset and value.

After AssetCommitment, ValueCommitment methods are added, the user can create asset, value commitment used in confidential transactions.

It closes #63 .
It closes #64 .


confidential/confidential.go
```diff
@@ -22,19 +22,19 @@ type UnblindOutputResult struct {
 }
 
 //NonceHash method generates hashed secret based on ecdh
-func NonceHash(ctx *secp256k1.Context, pubKey, privKey []byte) (*[32]byte, error) {
+func NonceHash(ctx *secp256k1.Context, pubKey, privKey []byte) ([32]byte, error) {
```
altafan, line 25: It would be more readable if  we name the return variables here, so we can avoid to return a `[32]byte{}` when an error occurs:
```suggestion
func NonceHash(ctx *secp256k1.Context, pubKey, privKey []byte) (result [32]byte, err error) {
```

confidential/confidential.go
```diff
@@ -91,3 +91,34 @@ func FinalValueBlindingFactor(input FinalValueBlindingFactorInput) ([32]byte, er
...
+func AssetCommitment(asset []byte, factor []byte) ([33]byte, error) {
+	ctx, _ := secp256k1.ContextCreate(secp256k1.ContextBoth)
+	defer secp256k1.ContextDestroy(ctx)
+
+	generator, err := secp256k1.GeneratorGenerateBlinded(ctx, asset, factor)
+	if err != nil {
+		return [33]byte{}, err
+	}
+
+	return secp256k1.GeneratorSerialize(ctx, generator), nil
```
altafan, line 105: You should be able to call the `Bytes` method of the `*Generator` type instead of calling the GeneratorSerialize binding:
```suggestion
	return generator.Bytes(), nil
```

confidential/confidential.go
```diff
@@ -91,3 +91,34 @@ func FinalValueBlindingFactor(input FinalValueBlindingFactorInput) ([32]byte, er
...
+	if err != nil {
+		return [33]byte{}, err
+	}
+
+	commit, err := secp256k1.Commit(ctx, factor, value, gen)
+	if err != nil {
+		return [33]byte{}, err
+	}
+
+	return secp256k1.CommitmentSerialize(ctx, commit)
```
altafan, line 123: same here, also the `*Commitment` type of the *go-secp256k1-zkp* lib has an available  `Bytes()` method:
```suggestion
	return commit.Bytes(), nil
```

confidential/confidential.go
sekulicd: Done

confidential/confidential.go
sekulicd: Done

confidential/confidential.go
sekulicd: Done

confidential/confidential.go
```diff
@@ -22,18 +22,18 @@ type UnblindOutputResult struct {
 }
 
 //NonceHash method generates hashed secret based on ecdh
-func NonceHash(ctx *secp256k1.Context, pubKey, privKey []byte) ([32]byte, error) {
+func NonceHash(ctx *secp256k1.Context, pubKey, privKey []byte) (result [32]byte, err error) {
 	_, publicKey, err := secp256k1.EcPubkeyParse(ctx, pubKey)
 	if err != nil {
-		return [32]byte{}, err
+		return result, err
```
altafan, line 28:
```suggestion
		return
```

confidential/confidential.go
```diff
@@ -22,18 +22,18 @@ type UnblindOutputResult struct {
...
 	}
 
 	_, ecdh, err := secp256k1.Ecdh(ctx, publicKey, privKey)
 	if err != nil {
-		return [32]byte{}, err
+		return result, err
```
altafan, line 33:
```suggestion
		return
```

confidential/confidential.go
```diff
@@ -22,18 +22,18 @@ type UnblindOutputResult struct {
...
 	}
 
-	result := sha256.Sum256(ecdh)
+	result = sha256.Sum256(ecdh)
 	return result, nil
```
altafan, line 37:
```suggestion
	return
```

confidential/confidential.go
```diff
@@ -93,32 +93,34 @@ func FinalValueBlindingFactor(input FinalValueBlindingFactorInput) ([32]byte, er
...
 	}
 
 	commit, err := secp256k1.Commit(ctx, factor, value, gen)
 	if err != nil {
-		return [33]byte{}, err
+		return result, err
 	}
 
-	return secp256k1.CommitmentSerialize(ctx, commit)
+	return commit.Bytes(), nil
```
altafan, line 125:
```suggestion
	result = commit.Bytes()
	return
```

confidential/confidential.go
```diff
@@ -93,32 +93,34 @@ func FinalValueBlindingFactor(input FinalValueBlindingFactorInput) ([32]byte, er
...
 //AssetCommitment method generates asset commitment
-func AssetCommitment(asset []byte, factor []byte) ([33]byte, error) {
+func AssetCommitment(asset []byte, factor []byte) (result [33]byte, err error) {
 	ctx, _ := secp256k1.ContextCreate(secp256k1.ContextBoth)
 	defer secp256k1.ContextDestroy(ctx)
 
 	generator, err := secp256k1.GeneratorGenerateBlinded(ctx, asset, factor)
 	if err != nil {
-		return [33]byte{}, err
+		return result, err
```
altafan, line 102:
```suggestion
		return
```

confidential/confidential.go
```diff
@@ -93,32 +93,34 @@ func FinalValueBlindingFactor(input FinalValueBlindingFactorInput) ([32]byte, er
...
 	}
 
-	return secp256k1.GeneratorSerialize(ctx, generator), nil
+	result = generator.Bytes()
+
+	return result, nil
```
altafan, line 107:
```suggestion
	return
```

confidential/confidential.go
```diff
@@ -93,32 +93,34 @@ func FinalValueBlindingFactor(input FinalValueBlindingFactorInput) ([32]byte, er
...
 //ValueCommitment method generates value commitment
-func ValueCommitment(value uint64, generator []byte, factor []byte) ([33]byte, error) {
+func ValueCommitment(value uint64, generator []byte, factor []byte) (result [33]byte, err error) {
 	ctx, _ := secp256k1.ContextCreate(secp256k1.ContextBoth)
 	defer secp256k1.ContextDestroy(ctx)
 
 	gen, err := secp256k1.GeneratorParse(ctx, generator)
 	if err != nil {
-		return [33]byte{}, err
+		return result, err
```
altafan, line 117:
```suggestion
		return
```

confidential/confidential.go
```diff
@@ -93,32 +93,34 @@ func FinalValueBlindingFactor(input FinalValueBlindingFactorInput) ([32]byte, er
...
 	}
 
 	commit, err := secp256k1.Commit(ctx, factor, value, gen)
 	if err != nil {
-		return [33]byte{}, err
+		return result, err
```
altafan, line 122:
```suggestion
		return
```

confidential/confidential.go
sekulicd: Done

confidential/confidential.go
sekulicd: Done

confidential/confidential.go
sekulicd: Done

confidential/confidential.go
sekulicd: Done

confidential/confidential.go
sekulicd: Done

confidential/confidential.go
sekulicd: Done

confidential/confidential.go
sekulicd: Done

confidential/confidential.go
sekulicd: Done

PR #78, opened 2020-06-03
Confidential elements to satoshi value format and viceversa

### LINKED ISSUE #66: [Confidential] Surjection proof
We need to expose a `SurjectionProof` method in our *confidential* package that expects the asset id of the output to blind, the output asset blinding factor, all the asset commitments and blinding factors of the *unblinded* output used as inputs in the transaction and a random 32-byte array seed and returns a surjection proof.

This method uses the params to initialize and generate a surjection proof with `secp256k1_surjectionproof_inititialize` and `secp256k1_surjectionproof_inititialize`, and finally returns the serialized raw bytes thanks to `secp256k1_surjectionproof_serialize`.

[This](https://github.com/vulpemventures/liquidjs-lib/blob/97673986c4233dee61f98dc59b5256ed2a8f313d/ts_src/confidential.ts#L133) can be used as reference implementation, while [here](https://github.com/vulpemventures/liquidjs-lib/blob/97673986c4233dee61f98dc59b5256ed2a8f313d/ts_src/psbt.ts#L786) you can see how this is used by the blinder role of the liquidjs-lib's *pset* module.

### LINKED ISSUE #75: [Confidential] Elements to satoshi value format and viceversa
We need to create in our `confidential` package 2 methods for converting an output value from satoshi (`uint64`) to the Elements confidential version (`[]byte`), and viceversa for unblinded values.

* `SatoshiToElementsValue(val uint64) ([9]byte, error)` expects a non-zero amount and converts it to raw bytes. [This](https://github.com/vulpemventures/go-elements/blob/master/transaction/transaction_test.go#L213) could simply be renamed and moved into the confidential package. Also in the pset package there is the same implementation of this function in the tests, please update both accordingly to use the new function.
* `ElementsToSatoshiValue(val []byte) (uint64, error)` expects a non-confidential amount in Elements format and converts it into satoshi amount. The length of non-confidential amount in Elements format is always **9** bytes, the first is the prefix `0x01` and the rest is the reverse bytes of the amount.

These functions should be tested together by checking that the result of converting a random uint64 value to Elements version with `SatoshiToElementsValue` and then back to Bitcoin version with `ElementsToSatoshiValue` matches the original one.

sekulicd:
Before there was no option to convert from Satoshi to Elements value and vice verse.

After the methods SatoshiToElementsValue/ElementsToSatoshiValue are added, it is possible to convert from Satoshi to Elements value and vice verse.

Note that new package 'common' is created, which contains files previously located in the transaction package.
This was done in order to resolve a circular dependency issue.

It closes #75  .


altafan:
@tiero do you think it is the case to merge the `common` and the `internal/bufferutil` packages into the `bufferutil` one?
At the end of the day the `ReverseBytes` function of the `internal/bufferutil` could be useful for users if they ever need to reverse a tx hash or whatever for any reason.
At the same time, It would be better to give a more meaningful name to the `common` package and `bufferutil` fits well IMO.

confidential/confidential.go
```diff
@@ -1,10 +1,18 @@
...
+	"bytes"
 	"crypto/sha256"
+	"encoding/binary"
+	"github.com/vulpemventures/go-elements/common"
+	"github.com/vulpemventures/go-elements/internal/bufferutil"
 	"github.com/vulpemventures/go-secp256k1-zkp"
 )
 
+const (
+	confidentialValue = 9
```
altafan, line 13: We should export this value, I would rename it like:
```suggestion
	ElementsUnconfidentialValueLength = 9
```

confidential/confidential.go
```diff
@@ -298,3 +306,25 @@ func SurjectionProof(input SurjectionProofInput) ([]byte, error) {
...
+		return
+	}
+	copy(result[:], append([]byte{unconfPrefix}, bufferutil.ReverseBytes(b.Bytes())...))
+
+	return
+}
+
+//ElementsToSatoshiValue method converts Elements value to Satoshi value
+func ElementsToSatoshiValue(val [9]byte) (result uint64, err error) {
+	reverseValueBuffer := [confidentialValue - 1]byte{}
```
altafan, line 324: Here the first thing to do is to ensure that `val` hash the correct `0x01` prefix, otherwise return a `invalid prefix` error
```suggestion
      if val[0] != byte(1) {
           err = errors.New("invalid prefix")
           return
      }
	reverseValueBuffer := [confidentialValue - 1]byte{}
```

confidential/confidential_test.go
```diff
@@ -413,3 +413,18 @@ func TestSurjectionProof(t *testing.T) {
 	}
 
 }
+
+func TestSatoshiToElementsValueAndBack(t *testing.T) {
+	var satoshi uint64 = 200000000
```
altafan, line 418: can we use a random number generator function?

confidential/confidential.go
```diff
@@ -298,3 +306,25 @@ func SurjectionProof(input SurjectionProofInput) ([]byte, error) {
 
 	return secp256k1.SurjectionProofSerialize(ctx, proof)
 }
+
+//SatoshiToElementsValue method converts Satoshi value to Elements value
+func SatoshiToElementsValue(val uint64) (result [9]byte, err error) {
```
altafan, line 311: Use the constant:
```suggestion
func SatoshiToElementsValue(val uint64) (result [ElementsUnconfidentialValueLength]byte, err error) {
```

confidential/confidential.go
```diff
@@ -298,3 +306,25 @@ func SurjectionProof(input SurjectionProofInput) ([]byte, error) {
...
+	if err = common.BinarySerializer.PutUint64(b, binary.LittleEndian, val); err != nil {
+		return
+	}
+	copy(result[:], append([]byte{unconfPrefix}, bufferutil.ReverseBytes(b.Bytes())...))
+
+	return
+}
+
+//ElementsToSatoshiValue method converts Elements value to Satoshi value
+func ElementsToSatoshiValue(val [9]byte) (result uint64, err error) {
```
altafan, line 323: use the constant:
```suggestion
func ElementsToSatoshiValue(val [ElementsUnconfidentialValueLength]byte) (result uint64, err error) {
```

confidential/confidential.go
sekulicd: Done

confidential/confidential.go
sekulicd: Done

confidential/confidential.go
sekulicd: Done

confidential/confidential.go
sekulicd: Done

tiero:
> @tiero @sekulicd do you think it is the case to merge the `common` and the `internal/bufferutil` packages into the `bufferutil` one?

I think users of go packages should be able to grasp its purpose from just the package's name. Common makes it difficult. See https://blog.golang.org/package-names and https://rakyll.org/style-packages/

> At the end of the day the `ReverseBytes` function of the `internal/bufferutil` could be useful for users if they ever need to reverse a tx hash or whatever for any reason.

I would keep it internally used only by the library and try to give high level interfaces to the user (eg. no need to revers). We could create a public package in a separate repository that does only this (if doesn't exist already)

confidential/confidential_test.go
sekulicd: Done

altafan:
>I think users of go packages should be able to grasp its purpose from just the package's name. Common makes it difficult. See https://blog.golang.org/package-names and https://rakyll.org/style-packages/

Yes, I agree with this in fact my purpose was avoiding to use the name `common` for a package.

> I would keep it internally used only by the library and try to give high level interfaces to the user (eg. no need to revers). We could create a public package in a separate repository that does only this (if doesn't exist already) and if it's really needed.

Ok, so what if we basically move the whole actual `common` package into `internal/bufferutil`? 
I mean, like the user of the lib does not need to reverse a byte slice, he shouldn't  also need to use the Serializer and Deserializer low-level types.

tiero:
> Ok, so what if we basically move the whole actual `common` package into `internal/bufferutil`?

I prefer this way. Makes sense to you? @sekulicd

sekulicd:
> > Ok, so what if we basically move the whole actual `common` package into `internal/bufferutil`?
> 
> I prefer this way. Makes sense to you? @sekulicd

In general project structure layout, in Golang, is bit gray area to me.
Its clear that the goal is to have simple, short packages that can be used by clients, but sometimes I would say, its hard to apply some architectural design pattern without having folder hierarchies.
Links you pasted in previous comments are well known and provides general guidance only. 

I think that project structure layout should vary based on the type of the project (Api, SDK, CLI etc).
In previous projects I was trying to follow recommendations from [here](https://github.com/golang-standards/project-layout) with some custom decisions.

As you know many important projects are written in Go and we can check and try to apply some patterns when starting new project.

In case of this project, which is sort of SDK, to be used by a various client I think we are on good track and that using internal package for code which is not meant to be used by client is good idea.

tiero:
Agreed @sekulicd

PR #79, opened 2020-06-04
Pset add blinder role

### LINKED ISSUE #45: [PSET] Add Blinder role
The blinder role is responsible for taking an **unsigned** partial transaction (thus it needs to check that every partial input has a empty `PartialSignature`) with unblinded outputs and proceeds to make them confidential by replacing their asset and values with commitments and by adding a nonce, and range and surjection proofs.
In this first version of the blinder role, we do not care about blinding issuance inputs since we do not support non-confidential issuance yet.

The process of blinding the outputs of a transaction is composed of the following steps:

## 1. Unblind tx inputs
First of all, we need to unblind every confidential utxo used as input of the transaction to get their asset/value commitments and blinding factors.
In order to accomplish this, every partial input must have a valid `WitnessUtxo` or `NonWitnessUtxo` set, so that we can determine if the utxo it refers to is confidential by checking if it has a range and a surjection proof.
Thus, to unblind all the confidential inputs we need the user to provide a list of valid blinding private keys.

```
for every partial input {
   1. get utxo from WitnessUtxo/NonWitnessUtxoField of partial input
   2. unblind utxo with respective blinding private key provided
   3. store unblinded value, asset, and blinding factors 
}
```

[Take a look](https://github.com/vulpemventures/liquidjs-lib/blob/master/ts_src/psbt.ts#L690) at how this step is implemented in the liquid-js library.

## 2. Generate blinding factors
The second step is to create asset and value blinding factors for every tx output, except for the fee out that must be kept non-confidential (this output is recognizable because it's the only one in a tx that has an empty script).
Every asset blinding factor is generated randomly by a strong *rng* function.
Every value blinding factor **except for one of the last output** is generated randomly by a strong *rng* function.
The value blinding factor of the last output is generated by calling the `FinalValueBlindingFactor` function, passing the previously unblinded input assets/values and blinding factors, and passing also the output values and the just created random asset/value blinidng factors:
```
for every output -1 (exclude the fee out!) {
   generate and store random asset blinding factors
}

for every output -2 (exclude fee and last output) {
   generate and store random value blinding factors
}

generate last value blinding factor with FinalValueBlindingFactorInput
add the last vbf to the array of value blinding factors
```

[Here](https://github.com/vulpemventures/liquidjs-lib/blob/master/ts_src/psbt.ts#L735) you can see how this step is implemented in the liquid-js library.

## 3 Blind tx outputs
Now that we are ready, it's time to effectively blind all tx outputs.
Then, for every output, we create the asset and value commitments from the unblinded asset/value and the respective blinding factors we created before.
We create an ephemeral random private key we need for generating the range proof. 
The public key derived from it is used as the output nonce.
Last, we create the range and surjection proofs, [check this out](https://github.com/vulpemventures/liquidjs-lib/blob/master/ts_src/psbt.ts#L762) for the implementation details for creating the proofs.
```
for every output {
   1. skip if the current output is the fee one
   2. create value and asset commitments for the current output
   3. create random private key
   4. set derived public key as the nonce of the current output
   5. create range proof with current output data, ephemeral privkey and current blinding pubkey
   6. create a random seed (entropy for surjection proof)
   7. create surjection proof with output data and random seed
}
```

### LINKED ISSUE #74: [Integration test] User can create, sign confidential inputs, blind confindetial outputs and broadcast the tx
it closes #91

sekulicd:
Before there was no option to calculate blind pset outputs.

After the method BlindOutputs is added, it is possible to blind outputs of PSET.

It closes #45  .

As per comment [here](https://github.com/vulpemventures/go-elements/pull/79#pullrequestreview-424585875) this will close #74 


pset/blinder.go
```diff
@@ -0,0 +1,314 @@
...
+
+func (b *blinder) unblindInputs() ([]confidential.UnblindOutputResult, error) {
+	ctx, _ := secp256k1.ContextCreate(secp256k1.ContextBoth)
+	defer secp256k1.ContextDestroy(ctx)
+	unblindOutputs := make([]confidential.UnblindOutputResult, 0)
+	for index, input := range b.pset.Inputs {
+		var prevout *transaction.TxOutput
+		if input.NonWitnessUtxo != nil {
+			//TODO: check if prevout should be created like below
+			prevout = input.NonWitnessUtxo.Outputs[index]
```
altafan, line 131: here we should use the unsigned tx input's `Index` instead of the actual index of the input in the transaction.
This sounds like a joke, but the `TxHash` and `Index` of a tx input actually refer to the specif output at index `Index` of the transaction with id `TxHash`:
```suggestion
			prevout = input.NonWitnessUtxo.Outputs[input.Index]
```

Suggestion relies on iterating over unsigned tx's inputs

pset/blinder.go
```diff
@@ -0,0 +1,314 @@
...
+			BlindingPubkey:      b.blindingPubkeys[outputIndex],
+			EphemeralPrivkey:    ephemeralPrivKey.Serialize(),
+			Asset:               outputAsset,
+			AssetBlindingFactor: outputAbfs[outputIndex],
+			ValueBlindFactor:    outVbf,
+			ValueCommit:         valueCommitment[:],
+			ScriptPubkey:        outputScript,
+			MinValue:            1,
+			Exp:                 0,
+			MinBits:             36,
```
altafan, line 273:
```suggestion
			MinBits:             52,
```

pset/blinder.go
```diff
@@ -0,0 +1,314 @@
...
+
+		if len(prevout.RangeProof) > 0 && len(prevout.SurjectionProof) > 0 {
+			val, err := secp256k1.CommitmentParse(ctx, prevout.Value)
+			if err != nil {
+				return nil, err
+			}
+			var commitmentValue secp256k1.Commitment
+			if val != nil {
+				commitmentValue = *val
+			}
```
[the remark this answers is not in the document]
sekulicd: secp256k1.CommitmentParse returns *Commitment.
ValueCommit(UnblindInput) is of type Commitment, since I need to dereference I wanted to prevent null pointer panic in case secp256k1.CommitmentParse returns nil.
I guess its a small chance that this can happen but wanted to be on the safe side.
Do you want me to remove it?

pset/blinder.go
```diff
@@ -0,0 +1,314 @@
...
+			unblindInput := confidential.UnblindInput{
+				EphemeralPubkey: b.blindingPubkeys[index],  //TODO: is this ok?
```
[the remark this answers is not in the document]
sekulicd: Done

pset/blinder.go
```diff
@@ -0,0 +1,314 @@
...
+				BlindingPrivkey: b.blindingPrivkeys[index], //TODO: is this ok?
+				Rangeproof:      prevout.RangeProof,
+				ValueCommit:     commitmentValue,
```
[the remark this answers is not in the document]
sekulicd: check [here](https://github.com/vulpemventures/go-elements/pull/79#discussion_r435305391)

pset/blinder.go
```diff
@@ -0,0 +1,314 @@
...
+		if len(outputScript) == 0 {
+			continue
+		}
+
+		randomSeed, err := generateRandomNumber(32)
+		if err != nil {
+			return err
+		}
+
+		//TODO: is it ok to use btcec?
```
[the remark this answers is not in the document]
sekulicd: Done

pset/blinder.go
```diff
@@ -0,0 +1,314 @@
...
+		if err != nil {
+			return err
+		}
+
+		b.pset.UnsignedTx.Outputs[outputIndex].Asset = assetCommitment[:]
+		b.pset.UnsignedTx.Outputs[outputIndex].Value = valueCommitment[:]
+
+		//TODO: should i calculate Nonce using confidential package?
+		b.pset.UnsignedTx.Outputs[outputIndex].Nonce = outputNonce.
+			SerializeUncompressed()
```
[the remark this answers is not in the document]
sekulicd: Done

pset/blinder.go
```diff
@@ -0,0 +1,314 @@
...
+		return err
+	}
+
+	unblindOutputs, err := b.unblindInputs()
+	if err != nil {
+		return err
+	}
+
+	outputValues := make([]uint64, 0)
+	//TODO check if looping over UnsignedTx's Outputs is correct?
```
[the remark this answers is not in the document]
sekulicd: Done

pset/blinder.go
sekulicd: Done

pset/blinder.go
```diff
@@ -0,0 +1,314 @@
...
+			"of outputs (fee excluded)")
+	}
+	return nil
+}
+
+func (b *blinder) unblindInputs() ([]confidential.UnblindOutputResult, error) {
+	ctx, _ := secp256k1.ContextCreate(secp256k1.ContextBoth)
+	defer secp256k1.ContextDestroy(ctx)
+	unblindOutputs := make([]confidential.UnblindOutputResult, 0)
+	for index, input := range b.pset.Inputs {
```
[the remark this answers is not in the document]
sekulicd: Done

pset/blinder.go
sekulicd: Done

pset/blinder.go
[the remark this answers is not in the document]
altafan: `secp256k1.CommitmentParse` should return a nil commitment only in case an error has occurred. Although I understand the concern, I think we should be good without this.

What's really annoying for me, is that we do not have a `CommitmentFromBytes` to avoid managing the creation and destruction of a context here. We'll fix this, if you could just  add a todo comment for this.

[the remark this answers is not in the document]
sekulicd: Done

sekulicd:
> It would be nice if the blinder could expect an argument with a rng function so that it uses the `generateRandomNumber` only as a fallback.

Can you elaborate more on this?

pset/blinder.go
```diff
@@ -0,0 +1,314 @@
...
+	)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (b *blinder) validate() error {
+	for _, input := range b.pset.Inputs {
```
altafan, line 104: we should also check that none of the inputs has a `PartialSignature`.

sekulicd: Done

pset/blinder.go
```diff
@@ -0,0 +1,314 @@
...
+
+	unblindOutputs, err := b.unblindInputs()
+	if err != nil {
+		return err
+	}
+
+	outputValues := make([]uint64, 0)
+	//TODO check if looping over UnsignedTx's Outputs is correct?
+	for _, output := range b.pset.UnsignedTx.Outputs {
+		var val [9]byte
```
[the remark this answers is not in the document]
altafan: @sekulicd

pset/blinder.go
```diff
@@ -106,6 +105,10 @@ func (b *blinder) validate() error {
 			return errors.New("all inputs must contain a non witness " +
 				"utxo or a witness utxo")
 		}
+
+		if len(input.PartialSigs) > 0 {
+			return errors.New("non of the inputs can't be partiali signed")
```
altafan, line 110: typos:
```suggestion
			return errors.New("inputs must not contain signatures")
```

pset/blinder.go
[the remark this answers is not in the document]
sekulicd: Done

pset/blinder.go
```diff
@@ -1,23 +1,25 @@
...
 import (
-	"crypto/rand"
 	"errors"
 	"github.com/btcsuite/btcd/btcec"
 	"github.com/vulpemventures/go-elements/confidential"
 	"github.com/vulpemventures/go-elements/transaction"
 	"github.com/vulpemventures/go-secp256k1-zkp"
 )
 
+type randomNumberGenerator func(n int) ([]byte, error)
```
altafan, line 11: we have 2 options here:
1. do not use any argument and then hardcode in our generateRandomNumber the `32` value
2. use an `interface{}` so the user is not forced to create his rng function with an `int` argument

I prefer number 1

pset/blinder.go
```diff
@@ -1,23 +1,25 @@
...
 	pset             *Pset
 	blindingPrivkeys [][]byte
 	blindingPubkeys  [][]byte
+	rng              randomNumberGenerator
 }
 
 // NewBlinder returns a new instance of blinder, if the passed Pset struct is
 // in a valid form, else an error.
-func NewBlinder(pset *Pset, blindingPrivkeys, blindingPubkeys [][]byte) (
+func NewBlinder(pset *Pset, blindingPrivkeys, blindingPubkeys [][]byte, rng func(n int) ([]byte, error)) (
```
altafan, line 22: use the defined type:
```suggestion
func NewBlinder(pset *Pset, blindingPrivkeys, blindingPubkeys [][]byte, rng randomNumberGenerator) (
```

pset/blinder.go
sekulicd: Please check now if this is ok now.

sekulicd:
I am trying to blind same PSET tx as in [this](https://github.com/vulpemventures/go-elements/blob/master/pset/pset_test.go#L303) test case, so there is one input and 3 outputs.
 My test is failing in this [line](https://github.com/vulpemventures/go-secp256k1-zkp/blob/a3b2b8c22053716c01c813042ebd91182a807f5c/pedersen.go#L246) cause vbl=4 != gbl=3.
vbl = input + output values = 1 input + 3 output
gbl = generatorBlindings = inputVbfs + outpuVbfs = 1 input + 2output

I was debugging mentioned JS test case and I see that method blindGeneratorBlindSum receives input with same count of vbl and gbl and it TC is passing 
Screenshot 2020-06-05 at 14 36 37

@altafan do you think that it can be issue with validation in [line](https://github.com/vulpemventures/go-secp256k1-zkp/blob/a3b2b8c22053716c01c813042ebd91182a807f5c/pedersen.go#L246) ?

altafan:
My guess is that you are including the value of the fee output within `values` while it should not.

pset/blinder.go
```diff
@@ -239,7 +239,7 @@ func (b *blinder) blindOutputs(
 	inputAgs [][]byte,
 	inputAbfs [][]byte,
 ) error {
-	for outputIndex, _ := range b.pset.Outputs {
+	for outputIndex, _ := range outputValues {
```
tiero, line 242: 🤯

tiero: didn't noticed mysel, i was trying to debug. Nice catch

sekulicd:
@altafan, test case is passing now and I can see confirmed blinded transaction in the explorer.
can you double-check if anything should be updated?

altafan:
Awesome!

LGTM, but I would add another test cases that uses confidential input instead of non confidential ones. Basically we need to do the same test, but instead of funding a bech32 address with the faucet, we fund a blech32 one.

This means that at the very beginning of the test we create a random private blinding key that we will use to then unblind the faucet output that we'll use as input of our test transaction. The public key derived, instead, is used to create the confidential blech32 address.

sekulicd:
> Awesome! @sekulicd
> 
> LGTM, but I would add another test cases that uses confidential input instead of non confidential ones. Basically we need to do the same test, but instead of funding a bech32 address with the faucet, we fund a blech32 one.
> 
> This means that at the very beginning of the test we create a random private blinding key that we will use to then unblind the faucet output that we'll use as input of our test transaction. The public key derived, instead, is used to create the confidential blech32 address.

Considering that this is the way to create blech32 address:
```
        privkey, err := btcec.NewPrivateKey(btcec.S256())
	if err != nil {
		t.Fatal(err)
	}
	pubkey := privkey.PubKey()

	publicKeyHashCompressed := btcutil.Hash160(pubkey.SerializeCompressed())
	fmt.Println(hex.EncodeToString(publicKeyHashCompressed))

	program := append(pubkey.SerializeCompressed(), publicKeyHashCompressed...)
	fmt.Println(hex.EncodeToString(program))

	blech32Addr := &address.Blech32{
		Prefix:  "el",
		Version: 0,
		Data:    program,
	}
```

How should I create scriptpubkey and signature, as we dont have option to create blech32 payment which would hold these two values like in case of p2wpkh for eg.:
```
changeScript := p2wpkh.Script
legacyScript := append(append([]byte{0x76, 0xa9, 0x14}, p2wpkh.Hash...), []byte{0x88, 0xac}...)
```

sekulicd:
> > Awesome! @sekulicd
> > LGTM, but I would add another test cases that uses confidential input instead of non confidential ones. Basically we need to do the same test, but instead of funding a bech32 address with the faucet, we fund a blech32 one.
> > This means that at the very beginning of the test we create a random private blinding key that we will use to then unblind the faucet output that we'll use as input of our test transaction. The public key derived, instead, is used to create the confidential blech32 address.
> 
> Considering that this is the way to create blech32 address:
> 
> ```
>         privkey, err := btcec.NewPrivateKey(btcec.S256())
> 	if err != nil {
> 		t.Fatal(err)
> 	}
> 	pubkey := privkey.PubKey()
> 
> 	publicKeyHashCompressed := btcutil.Hash160(pubkey.SerializeCompressed())
> 
> 	program := append(pubkey.SerializeCompressed(), publicKeyHashCompressed...)
> 
> 	blech32Addr := &address.Blech32{
> 		Prefix:  "el",
> 		Version: 0,
> 		Data:    program,
> 	}
> ```
> 
> How should I create scriptpubkey and signature, as we dont have option to create blech32 payment which would hold these two values like in case of p2wpkh for eg.:
> 
> ```
> changeScript := p2wpkh.Script
> legacyScript := append(append([]byte{0x76, 0xa9, 0x14}, p2wpkh.Hash...), []byte{0x88, 0xac}...)
> ```

UPDATE:
@altafan, for the changeScript and legacyScript i took bellow values:
```
blech32ScriptPubKey := append([]byte{0x00, 0x14}, publicKeyHashCompressed...)
legacyScript := append(append([]byte{0x76, 0xa9, 0x14}, publicKeyHashCompressed...), []byte{0x88, 0xac}...)
```
Please advise if this is correct.

Assuming that this is correct I am trying to add WitnessUtxo:
`updater.AddInWitnessUtxo`
but considering that Utxo is now confidential, as we funded blech32 address, I need to add Surejction and Range proof.

Since Utxo doesnt contain Surjection and Range proof, idea is to fetch full transaction.
In order to do this I took txId from utxo and invoked:
`/tx/{txId}/hex `

I receive full transaction in raw bytes which I am able to decode using elements-cli, now the problem is that the method, I assumed, should be used to decode raw tx:
`transaxction.NewTxFromBuffer`
is throwing error and I am not able to fetch previous transaction from which I am planning to take Surjection and Range proof that are needed when unblinding inputs.
Please advise if I missed something.

altafan:
I was not intending to modify the running integration test by using a confidential input, but rather to create a new one.

Said that, the confidential address format, blech32 this case, is just a way so that I, receiver of a transaction, can pass my blinding pubkey to the sender in order to him being able to blind my outputs with a key that I'm aware of.

For a confidential output, thus, the output script does not change at all compared to the unconfidential one, so you should not be worried it.

The `WitnessUtxo` field of a partial input is of type `transaction.TxOutput` that already contains the field for range and surjection proofs.
If you fetch a utxo that does not have proofs it may mean that it's not a confidential utxo at all? 

In general the `NonWitnessUtxo` is used when managing utxos locked by p2pkh script, just to let you know.

PR #86, opened 2020-06-08
Payment support for confidential addresses

### LINKED ISSUE #83: [Payment] Support for base58 confidential addresses
At the moment we do not support the blinding key when creating a legacy base58 address. We need to add the blinding key when encoding and extract it when decoding

### LINKED ISSUE #85: [Payment] Support for blech32 confidential addresses
Taking in consideration the changes to do in #83, ~~also the `WitnessPubKeyHash` and the `WitnessScriptHash` function should return a confidential address by default if the payment's `BlindingKey` is not nil.~~
EDIT: we need to create 2 new methods `ConfidentialWitnessPubKeyHash` and `ConfidentialWitnessScriptHash` that check that the payment has a valid blinding pubkey and return the confidential address.

Here we can take advantage of the `ToBlech32` function of the `address` package that already takes care of encoding all the data in the correct way to return a valid confidential address. In case of doubts on how to use this function, the tests are a good place where to see how work with `ToBlech32`.

[This test](https://github.com/vulpemventures/liquidjs-lib/blob/97673986c4233dee61f98dc59b5256ed2a8f313d/test/fixtures/p2wpkh.json#L131) from liquid js whould be replicated for confidential p2wpkh address, while [this other one](https://github.com/vulpemventures/liquidjs-lib/blob/97673986c4233dee61f98dc59b5256ed2a8f313d/test/fixtures/p2wsh.json#L99) should be present for confidential p2wsh.

sekulicd:
Before there was no option to create p2pkh and p2sh confidential addresses using payment package.

After the methods ConfidentialPubKeyHash/ConfidentialScriptHash are added, it is possible to convert create confidential p2pkh and p2sh addresses.

It closes #83 .
It closes #85.


payment/payment_test.go
```diff
@@ -111,5 +117,59 @@ func TestFromPublicKeys(t *testing.T) {
...
+	}
+
+	payment := payment.FromPublicKey(pubKey1, &network.Liquid, pubKey2)
+	assert.Equal(t, expected, payment.ConfidentialPubKeyHash())
+}
+
+func TestPaymentConfidentialScriptHash(t *testing.T) {
+	expected := "VJLCUu2hpcjPaTGMnANXni8wVYjsCAiTEznE5zgRZZyAWXE2P6rz6Dvph" +
+		"BHSn7iz4w9sLb3mFSHGJbte"
+	scriptHash, err := hex.DecodeString(
```
altafan, line 152: we should use a script here instead of its hash. This wrongly led you to use the `Script` instead of the `Hash` into the `ConfidentialScriptHash`, I guess.

From the reference test vector, we should translate the [script](https://github.com/vulpemventures/liquidjs-lib/blob/97673986c4233dee61f98dc59b5256ed2a8f313d/test/fixtures/p2sh.json#L83) in ASM version to the hex encoding, so we can convert it to bytes in the test using `hex`.
The proper script should be:
```
a9149f840a5fc02407ef0ad499c2ec0eb0b942fb008687
```

payment/payment_test.go
```diff
@@ -111,5 +118,129 @@ func TestFromPublicKeys(t *testing.T) {
...
+		"9f840a5fc02407ef0ad499c2ec0eb0b942fb0086")
+	if err != nil {
```
altafan, line 155: Since this is an hardcoded value, you can skip checking the error here

payment/payment_test.go
```diff
@@ -111,5 +118,129 @@ func TestFromPublicKeys(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	script := append(append([]byte{txscript.OP_HASH160, 0x14}, scriptHash...),
```
altafan, line 159: Why not using directly the encoded script `a9149f840a5fc02407ef0ad499c2ec0eb0b942fb008687` instead of constructing them from the scriptHash?

payment/payment.go
```diff
@@ -95,7 +125,20 @@ func FromScript(script []byte, net *network.Network) (*Payment, error) {
...
+	if script[0] == txscript.OP_HASH160 {
+		scriptHash = append(scriptHash, script[2:len(script)-1]...)
+	}
+	if script[len(script)-1] == txscript.OP_CHECKMULTISIG {
+		scriptHash = hash160(script)
+	}
+
+	redeem := &Payment{Network: tmpNet, Hash: scriptHash, Script: script,
+		BlindingKey: blindingKey}
 	return FromPayment(redeem)
```
altafan, line 142: here we should not call the FromPayment but rather returning the payment instance:
```suggestion
	return &Payment{
	   Network: tmpNet,
	   Hash: scriptHash,
	   Script: script,
	   BlindingKey: blindingKey,
	}, nil
```

payment/payment.go
```diff
@@ -71,21 +81,41 @@ func FromPublicKeys(pubkeys []*btcec.PublicKey, nrequired int, net *network.Netw
 		return nil, err
 	}
 
-	return FromScript(multiSigScript, tmpNet)
+	return FromScript(multiSigScript, tmpNet, blindingKey)
```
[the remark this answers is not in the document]
sekulicd: Done

payment/payment.go
```diff
@@ -71,21 +81,41 @@ func FromPublicKeys(pubkeys []*btcec.PublicKey, nrequired int, net *network.Netw
...
+		payment.Redeem,
+		payment.Script,
+		payment.WitnessHash,
+	}
 	witnessHash := sha256.Sum256(redeem.Script)
-	return &Payment{payment.Network, payment.PublicKey, hash160(redeem.Script), payment.BlindingKey, redeem, payment.Script, witnessHash[:]}, nil
+	return &Payment{
+		payment.Network,
+		payment.PublicKey,
+		payment.Hash,
```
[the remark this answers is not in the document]
sekulicd: Done

payment/payment.go
```diff
@@ -95,7 +125,20 @@ func FromScript(script []byte, net *network.Network) (*Payment, error) {
...
+	scriptHash := make([]byte, 0)
+	if script[0] == txscript.OP_0 {
+		scriptHash = append(scriptHash, script[2:]...)
+	}
+	if script[0] == txscript.OP_HASH160 {
+		scriptHash = append(scriptHash, script[2:len(script)-1]...)
+	}
+	if script[len(script)-1] == txscript.OP_CHECKMULTISIG {
+		scriptHash = hash160(script)
+	}
```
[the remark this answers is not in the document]
sekulicd: Done

payment/payment_test.go
sekulicd: Done

payment/payment.go
sekulicd: Done

payment/payment_test.go
sekulicd: Done

payment/payment.go
```diff
@@ -134,6 +224,27 @@ func (p *Payment) WitnessPubKeyHash() (string, error) {
 	return addr, nil
 }
 
+// ConfidentialWitnessPubKeyHash is a method of the Payment struct to derive
+//a confidential base58 p2wpkh address
```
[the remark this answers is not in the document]
sekulicd: Done

payment/payment.go
```diff
@@ -152,6 +263,27 @@ func (p *Payment) WitnessScriptHash() (string, error) {
 	return addr, nil
 }
 
+// ConfidentialWitnessScriptHash is a method of the Payment struct to derive
+//a confidential base58 p2wsh address
```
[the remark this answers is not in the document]
sekulicd: Done

payment/payment.go
```diff
@@ -109,6 +157,27 @@ func (p *Payment) PubKeyHash() string {
 	return addr
 }
 
+// ConfidentialPubKeyHash is a method of the Payment struct to derive a
+//base58 confidential p2pkh address
```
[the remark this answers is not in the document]
sekulicd: Done

payment/payment.go
```diff
@@ -119,6 +188,27 @@ func (p *Payment) ScriptHash() (string, error) {
 	return addr, nil
 }
 
+// ConfidentialScriptHash is a method of the Payment struct to derive a
+//base58 confidential p2sh address
```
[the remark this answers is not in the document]
sekulicd: Done

payment/payment.go
```diff
@@ -189,7 +189,7 @@ func (p *Payment) ScriptHash() (string, error) {
 }
 
 // ConfidentialScriptHash is a method of the Payment struct to derive a
-//base58 confidential p2sh address
+//blech32 confidential p2sh address
```
tiero, line 192: here it's correct the comment `base58` as was before

payment/payment.go
```diff
@@ -158,7 +158,7 @@ func (p *Payment) PubKeyHash() string {
 }
 
 // ConfidentialPubKeyHash is a method of the Payment struct to derive a
-//base58 confidential p2pkh address
+//blech32 confidential p2pkh address
```
tiero, line 161: revert to `base58` in comment

payment/payment.go
sekulicd: Done

payment/payment.go
sekulicd: Done

PR #87, opened 2020-06-09
Issuance: generation of entropy, asset hash and reissuance token hash

### LINKED ISSUE #80: [Transaction] Method for generating issuance asset entropy
In order to support the non-confidential issuance of new assets, we need to define some methods for the `*TxIssuance` type in the transaction package for generating the asset entropy and calculating the asset hash and reissuance token from it.

A **non-confidential** issuance transaction is a transaction that contains:
* one input that includes a `TxIssuance`, that describes the entropy of the asset, its issuance amount and the reissuance token amount
* one output for the issuance amount
* one output for the reissuance token amount (if > 0)
Both the outputs are, of course, locked with the script of the minter of the asset.

We need to define a `GenerateAssetEntropy` function for `*TxIssuance` type that generates the entropy of an asset from a tx hash, an input index (this is the index of the issuance transaciton, not the prevout index!) and the hash of a Ricardian contract that represents the asset itself.
[Here](https://github.com/vulpemventures/go-elements/issues/26#issuecomment-638228277) you can see an example of a contract.

It would be better if the sha256 hash of the contract was done by our `GenerateAssetEntropy` so we define a type structure for the contract like the following and the user is aware of how to properly create a contract:

```
type Entity struct {
   Domain string
}

type Contract struct {
   Name string
   Ticker string
   Version uint
   Precision uint
   Entity Entity
}
```

The function should just return an error if any occur, otherwise it has the side effect of changing the `AssetEntropy` field of the issuance with the one generated.
This should be the function signature:

```
func (*TxIssuance) GenerateAssetEntropy(inTxHash []byte, inTxIndex, contract Contract) error {}
```

[Here](https://github.com/ElementsProject/libwally-core/blob/master/src/transaction.c#L2748) you can find the reference implementation for this function, below a quick explanation of the code:
* step 1: sha256 double hash of the serialized `[txHash, uint32LE(inputIndex)]`
* step 2: **midstate** sha256 hash of serialized `[stepOneResult, sha256(contract)]`

### LINKED ISSUE #81: [Transaction] Method for calculating asset hash
We need to define a `GenerateAssetHash` for the `*TxIssuance` type that calculates the hash of the asset to issue as the sha256 hash of the serialization of the entropy of the issuance and a `0x00..00` 32-byte array.
[Here](https://github.com/ElementsProject/libwally-core/blob/master/src/transaction.c#L2805) you can find the reference implementation.

### LINKED ISSUE #82: [Transaction] Method for calculating reissuance token
We need to define a `GenerateReissuanceToken` method for the `*TxIssuance` type that expects a flag argument that can assume only the values `0` or `1` and calculates from it and from the entropy of the issuance type the hash of the reissuance token as the sha256 hash of the serialized entropy and the `0x??..00` 32-byte array. `??` is actually the given flag + 1.
[Here](https://github.com/ElementsProject/libwally-core/blob/master/src/transaction.c#L2816) you can see the reference libwally's implementation.

altafan:
This adds some methods to the `*TxIssuance` type in order to:
* generate an issuance entropy from the given input's prev hash and index, and from an asset contract (can be not defined)
* generate the hash of the asset to issue
* generate the hash of the token for reissuing the asset in the future

This closes #80.
This closes #81.
This closes #82.

Note:
A new *issuance.go* file is added for collecting the definition of the `TxIssuance` type (moved here from *transaction.go) and the methods of its pointer.


transaction/issuance.go
```diff
@@ -0,0 +1,128 @@
...
+	"encoding/json"
+	"errors"
+
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/vulpemventures/fastsha256"
+	"github.com/vulpemventures/go-elements/confidential"
+	"github.com/vulpemventures/go-elements/internal/bufferutil"
+)
+
+const ()
```
tiero, line 13: this?

altafan: misprint

transaction/issuance.go
```diff
@@ -0,0 +1,126 @@
...
+// TxIssuance defines the type for Issuance field in TxInput
+type TxIssuance struct {
+	AssetBlindingNonce []byte
+	AssetEntropy       []byte
+	AssetAmount        []byte
+	TokenAmount        []byte
+}
+
+// NewTxIssuance returns a new empty issuance instance
+func NewTxIssuance(assetAmount, tokenAmount uint64) (*TxIssuance, error) {
```
tiero, line 36: We should pass the precision as parameter here also and then we would make to check in the GenerateEntropy method that the asset amount provided and precision together are correct.

transaction/issuance.go
```diff
@@ -0,0 +1,126 @@
...
+	}, nil
+}
+
+// GenerateEntropy generates the entropy from which the hash of the asset and
+// of the reissuance token are calculated
+func (issuance *TxIssuance) GenerateEntropy(inTxHash []byte, inTxIndex uint32, contract *IssuanceContract) error {
+	if len(inTxHash) != 32 {
+		return errors.New("invalid tx hash length")
+	}
+
```
tiero, line 65: we should check the provided contract matches the asset amount provided

altafan: the contract does not have information about the amount of the asset

transaction/issuance.go
altafan: do you know if the precision affects also the token amount or just the asset amount?

transaction/issuance.go
tiero: the contract no, but the TxIssuance struct yes.

transaction/issuance.go
tiero: I guess that the reissuance tokens could be maximum 21M*10**8 and not more/less. But thats' a good question

tiero: Screen Shot 2020-06-09 at 4 04 05 PM

@altafan

tiero:
> An issuance will be blinded by the blinder roel of the pset package, because this is a part of the blinding process that it actually misses.

Do you mean that the TxIssuance could not be treated as a common output and need more work on the blinder role?

altafan:
The TxIssuance is a part of a transaction input. It contains asset amount and token amount that are also add in the transaction as outputs, so that they can be used as utxo in future txs.

So, when blinding a transaction containing an issuance it is necessary to produce commitments and proof for the outputs of the transaction, but also for the input containing the issuance.

I was trying to say that this is something that the blinder has to care about when blinding the transaction. `BlindOutputs` may be an incorrect name after all.. maybe we will change it in `BlindTransaction` since it will potentially blind also inputs if they contain any issuance.

transaction/issuance.go
altafan: Consider that the fields of the `TxIssuance` type are just the ones that will be encoded in the transaction. Note that the eventual range proofs for confidential asset and token are not part of this type because they are encoded together with the witnesses and treated separatedly.

I agree that it would be nice to have the precision here, but I don't think we should add a non-encoded field to the TxIssuance type.

tiero:
> The TxIssuance is a part of a transaction input. It contains asset amount and token amount that are also add in the transaction as outputs, so that they can be used as utxo in future txs.
> 
> So, when blinding a transaction containing an issuance it is necessary to produce commitments and proof for the outputs of the transaction, but also for the input containing the issuance.
> 
> I was trying to say that this is something that the blinder has to care about when blinding the transaction. `BlindOutputs` may be an incorrect name after all.. maybe we will change it in `BlindTransaction` since it will potentially blind also inputs if they contain any issuance.

Thanks for explanation @altafan

transaction/issuance.go
tiero: I'm not saying we should add the field, but we should be able to check somehow tha the suer si providingi the right asset amount and the right precision. Maybe we should do later on a convenience high level function that call these?

tiero: @altafan

altafan:
@tiero I pushed the changes for the precision, can you review?

PR #98, opened 2020-06-11
Transaction method for–adding a non confidential issuance to a transaction

### LINKED ISSUE #26: [unconfidential] Issue an asset
An user should be able to make a raw issuance unblinded defining initial supply, re-issuance tokens and so on

### LINKED ISSUE #88: [Integration test] User can issue an unblinded asset.
This should close #26

### LINKED ISSUE #94: [PSET] Method for adding a non-confidential issuance to a transaction
We need to add a new `AddIssuance` method ~~to the `*Transaction` type of the *transaction* package.~~
EDIT-2: to the `*Updater` type of the *pset* package.

This new function will have the following signature:
```
func (u *Updater) AddIssuance(arg AddIssuanceArg) (error)
```

It expects a single argument defined like:

```
type AddIssuanceArg struct {
   Precision uint,
   Contract *IssuanceContract,
   AssetAmount float64,
   TokenAmount float64,
   AssetAddress string,
   TokenAddress string,
   TokenFlag uint,
}
```

About the fields of this type, it is important to note that:
* the asset and token amount are expressed in *BTC* rather than *Sat*. After instantiating a new issuance object, the amounts in satoshis can be taken from the latter.
* the function using these args will add some outputs to the transaction (as better explained below), so it will need to get the correct scripts from the provided addresses.

Following there are the steps that the new function have to do to correctly add an unblinded issuance to a transaction that **must** contain at least 1 input:

1) Check that transaction contains at least one input
2) Instantiate a new *TxIssuanceExtended.
3)
   a)  Check that AssetAmount and AssetAddress are defined
   b) Check that TokenAmount and TokenAddress are defined if TokenAmount is > 0
4) Get the index of the first input of the transaction that has empty issuance
5) Generate entropy, asset (and token if defined) using the selected input hash and index and the provided Contract
6) Add the issuance to the designed input
7)
   a) Add an output to the transaction with asset amount, asset hash and script derived from asset address
   b) Optionally add an output with token amount, token hash and script derived from token address if the provided TokenAmount is > 0
~~8) Finally, return the index of the selected input containing the issuance, or an error if any occurs~~

EDIT-1:
I added a new field `TokenFlag uint` to the `AddIssuanceArg` type. This will be used when generating the reissuance token hash, and its values can be `0` (non confidential issuance), or `1` (confidential issuance).
In step *7b* we should also enforce that this flag has a valid value, and we should declare 2 constants `NonConfidentialReissuanceTokenFlag` and `ConfidentialReissuanceTokenFlag` for the valid values so the user can use them.

EDIT-2:
The right place for this function is the **pset** package, specifically, the updater role should be responsible for "adding an issuance" to a transaction, as simple as it may be said in this way.
The steps remain the same except for the fact that we don't need to return the issuance input index anymore, thus we return an error if occurring.
This now depends on #100, in particular, we need to able to add new outputs to the partial transaction.

sekulicd:
Before there was no option to add non-confidential issuance to the transaction.

After the method AddIssuance is added, it is possible to add issuance to the transaction.
In order to calculate asset/token output script, two new public methods are added in the address package:
ToOutputScript - calculates output script
DecodeType - returnes address type.

It closes #94.
It closes #88.
It closes #26.


transaction/transaction.go
```diff
@@ -73,6 +78,106 @@ func NewTxInput(hash []byte, index uint32) *TxInput {
...
+	AssetAddress        string
+	TokenAddress        string
+	ReissuanceTokenFlag uint
+	Net                 network.Network
+}
+
+//AddIssuance adds non-confidential issuance to the transaction
+func (tx *Transaction) AddIssuance(arg AddIssuanceArg) (uint32, error) {
+	if len(tx.Inputs) == 0 {
+		return 0, errors.New("transaction must contain at least one input ")
```
altafan, line 95: return `-1` when returning error and remove space in error message:
```suggestion
		return 0, errors.New("transaction must contain at least one input")
```
This implies changing the return value from `uint32` to `int`

transaction/transaction.go
```diff
@@ -73,6 +78,106 @@ func NewTxInput(hash []byte, index uint32) *TxInput {
...
+	}
+
+	if arg.AssetAmount == 0 {
+		return 0, errors.New("asset amount must be greater then 0")
+	}
+
+	if arg.AssetAddress == "" {
+		return 0, errors.New("asset address must be provided")
```
altafan, line 103:
```suggestion
		return 0, errors.New("destination address for issued asset must be provided")
```

transaction/transaction.go
```diff
@@ -73,6 +78,106 @@ func NewTxInput(hash []byte, index uint32) *TxInput {
...
 }
 
+type AddIssuanceArg struct {
+	Precision           uint
+	Contract            *IssuanceContract
+	AssetAmount         float64
+	TokenAmount         float64
+	AssetAddress        string
+	TokenAddress        string
+	ReissuanceTokenFlag uint
```
altafan, line 88: Let's call this just `TokenFlag`

transaction/transaction.go
```diff
@@ -73,6 +78,106 @@ func NewTxInput(hash []byte, index uint32) *TxInput {
...
+	if arg.AssetAmount == 0 {
+		return 0, errors.New("asset amount must be greater then 0")
+	}
+
+	if arg.AssetAddress == "" {
+		return 0, errors.New("asset address must be provided")
+	}
+
+	if arg.TokenAmount > 0 && arg.TokenAddress == "" {
+		return 0, errors.New("token address must be provided")
```
altafan, line 107:
```suggestion
		return 0, errors.New("destination address for reissuance token must be provided")
```

Also, in case the TokenAmount is defined, we should check that the TokenFlag has a valid value

transaction/transaction.go
```diff
@@ -73,6 +78,106 @@ func NewTxInput(hash []byte, index uint32) *TxInput {
...
+	}
+
+	issuanceExt, err := NewTxIssuance(arg.AssetAmount, arg.TokenAmount,
```
altafan, line 110: call this `issuance` please:
```suggestion
	issuance, err := NewTxIssuance(arg.AssetAmount, arg.TokenAmount,
```

transaction/transaction.go
```diff
@@ -73,6 +78,106 @@ func NewTxInput(hash []byte, index uint32) *TxInput {
...
+		return 0, err
+	}
+	tx.Inputs[inputIndex].Issuance = &issuanceExt.TxIssuance
+
+	assetHash, err := issuanceExt.GenerateAsset()
+	if err != nil {
+		return 0, err
+	}
+
+	assetAmount, err := confidential.SatoshiToElementsValue(
```
altafan, line 136: you can avoid doing this since the issuance.TxIssuance.AssetAmount is already in the elements format ([]byte)

transaction/transaction.go
```diff
@@ -73,6 +78,106 @@ func NewTxInput(hash []byte, index uint32) *TxInput {
...
+		return 0, err
+	}
+	script, err := address.ToOutputScript(arg.TokenAddress, arg.Net)
+	if err != nil {
+		return 0, err
+	}
+
+	output := &TxOutput{
+		Asset:  assetHash,
+		Value:  assetAmount[:],
```
altafan, line 149: use amount in the issuance:
```suggestion
		Value:  issuance.TxIssuance.AssetAmount,
```

transaction/transaction.go
```diff
@@ -73,6 +78,106 @@ func NewTxInput(hash []byte, index uint32) *TxInput {
...
+	if err != nil {
+		return 0, err
+	}
+
+	var inputIndex uint32
+	var inputHash []byte
+	for i, input := range tx.Inputs {
+		if input.Issuance == nil {
+			inputIndex = uint32(i)
+			inputHash = input.Hash
```
altafan, line 121: can we also break the loop?

altafan:
@sekulicd I just left a comment in the task explaining how we need to change the behavior of this metod to allow using it into the pset package in the future.

altafan:
For the integration test, we actually need to add some methods to the updater role of the pset package in order to correctly add the issuance outputs to the partial transaction.
After these tasks are defined, I would suggest including them in this PR so we are sure to commit code that is working correctly into master.

transaction/transaction.go
```diff
@@ -73,6 +78,106 @@ func NewTxInput(hash []byte, index uint32) *TxInput {
...
+
+	issuanceExt, err := NewTxIssuance(arg.AssetAmount, arg.TokenAmount,
+		arg.Precision)
+	if err != nil {
+		return 0, err
+	}
+
+	var inputIndex uint32
+	var inputHash []byte
+	for i, input := range tx.Inputs {
```
altafan, line 118: considered the updates made to task, we should range over `u.Pset.UnsignedTx.Inputs`

transaction/transaction.go
```diff
@@ -73,6 +78,106 @@ func NewTxInput(hash []byte, index uint32) *TxInput {
...
+	if err != nil {
+		return 0, err
+	}
+
+	output := &TxOutput{
+		Asset:  assetHash,
+		Value:  assetAmount[:],
+		Script: script,
+	}
+	tx.AddOutput(output)
```
altafan, line 152: we should add the output by calling `u.AddOutput(output)`

transaction/transaction.go
```diff
@@ -73,6 +78,106 @@ func NewTxInput(hash []byte, index uint32) *TxInput {
...
+		if err != nil {
+			return 0, err
+		}
+
+		output := &TxOutput{
+			Asset:  tokenHash,
+			Value:  tokenAmount[:],
+			Script: script,
+		}
+		tx.AddOutput(output)
```
altafan, line 175: same here

address/address.go
```diff
@@ -231,3 +245,202 @@ func ToBlech32(bl *Blech32) (string, error) {
...
+//DecodeType returns address type
+func DecodeType(address string, net network.Network) (int, error) {
+	if isBlech32(address, net) {
+		return handleBlech32(address, net)
+	}
+	if isBech32(address, net) {
+		return handleBech32(address, net)
+	}
+	return handleBase58(address, net)
+}
```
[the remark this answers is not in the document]
sekulicd: Done

payment/payment_test.go
```diff
@@ -237,3 +251,223 @@ func TestConfidentialWitnessScriptHash(t *testing.T) {
...
+func TestDecodeAddressTypeConfidentialP2Pkh(t *testing.T) {
+	privKey1, err := btcec.NewPrivateKey(btcec.S256())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	privKey2, err := btcec.NewPrivateKey(btcec.S256())
+	if err != nil {
+		t.Fatal(err)
+	}
```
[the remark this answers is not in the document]
sekulicd: Done

transaction/transaction.go
sekulicd: Done

transaction/transaction.go
sekulicd: Done

transaction/transaction.go
sekulicd: Done

transaction/transaction.go
sekulicd: Done

transaction/transaction.go
sekulicd: Done

transaction/transaction.go
sekulicd: Done

transaction/transaction.go
sekulicd: Done

transaction/transaction.go
sekulicd: Done

transaction/transaction.go
sekulicd: Done

pset/updater.go
```diff
@@ -380,29 +386,34 @@ type AddIssuanceArg struct {
...
-		return 0, errors.New("destination address for reissuance token " +
+		return errors.New("destination address for reissuance token " +
 			"must be provided")
 	}
 
+	if arg.TokenFlag != NonConfidentialReissuanceTokenFlag &&
+		arg.TokenFlag != ConfidentialReissuanceTokenFlag {
+		return errors.New("token flag must be 0 or 1")
+	}
+
```
altafan, line 412: maybe it is the case to perform these checks only when the token amount is defined? 
```suggestion
	if arg.TokenAmount > 0 {
	        if arg.TokenAddress == "" {
	            return errors.New("destination address for reissuance token " +
			"must be provided")
	        }
	        if arg.TokenFlag != NonConfidentialReissuanceTokenFlag &&
		    arg.TokenFlag != ConfidentialReissuanceTokenFlag {
		    return errors.New("token flag must be 0 or 1")
		}
	}
```
This way we would just ignore token params when its amount is not defined

sekulicd: Done

pset/updater_test.go
```diff
@@ -77,3 +85,132 @@ func TestUpdaterAddOutput(t *testing.T) {
 	assert.Equal(t, 1, len(updater.Upsbt.UnsignedTx.Outputs))
 	assert.Equal(t, 1, len(updater.Upsbt.Outputs))
 }
+
+func TestCreateAddIssuanceSignAndBroadcast(t *testing.T) {
```
tiero, line 89: I do believe that this  e2e test belongs o pset_test rather than here.

tiero:
It seems it could close #88 could you edit the PR description to reflect this? @altafan

pset/updater_test.go
sekulicd: Done

PR #99, opened 2020-06-11
Slip77 support

### LINKED ISSUE #76: [Slip77] Type definition and factory functions
We should create a new package `slip77` containing the following type feinition:

```
type Slip77 struct {
   MasterKey []byte
   Extra []byte
}
```

These fields are used along with a provided output script to derive a blinding key pair in a deterministic way. This is really useful when blinding a transaction since the user does not need to create/remember every blinding key pair, but instead he just needs to provide a master key or a seed to subsequently derive all needed blinding key pairs from any tx output script.

Given that, we need to create a `FromMasterKey` factory function that expects a key as a byte slice and returns a pointer to the `Slip77` type. This method, thus, simply instantiates a new variable of type `Slip77` , sets the `MasterKey` for it and returns the reference to the variable.

We should also expose a `FromSeed` factory function that derives the master key from a bip39 seed and then returns the pointer to the Slip77 instance as the previous function. We could use [this implementation](https://github.com/tyler-smith/go-bip39) of bip39, it's listed among the implementations directly [in the bip](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#other-implementations).

### LINKED ISSUE #77: [Slip77] Derive blinding keys from output script
We need to define a `Derive` function for the `*Slip77` type that expects an output script and returns a couple of private, public blinding keys.
It basically creates the sha256 hmac for the given master key and output script and it to calculate the point on the ellyptic curve from this scalar.
[This](https://github.com/vulpemventures/slip77/blob/master/ts_src/index.ts#L75) can be used as reference implementation.

altafan:
This adds a new `slip77` package to the library that enables the user to derive blinding keys from a master key or a bip39 seed, and an output script following the SLIP-0077 specification.

This closes #76.
This closes #77.

@tiero please, review this.

### DIFF
--- a/slip77/slip77.go
+++ b/slip77/slip77.go
@@ -1,0 +1,69 @@
+package slip77
+
+import (
+	"crypto/hmac"
+	"crypto/sha256"
+	"crypto/sha512"
+	"errors"
+
+	"github.com/btcsuite/btcd/btcec"
+)
+
+var (
+	domain = []byte("Symmetric key seed")
+	label  = []byte("SLIP-0077")
+	prefix = byte(0)
+)
+
+type Slip77 struct {
+	MasterKey []byte
+}
+
+// FromMasterKey sets the provided master key to the returned instance of Slip77
+func FromMasterKey(masterKey []byte) (*Slip77, error) {
+	if masterKey == nil || len(masterKey) <= 0 {
+		return nil, errors.New("invalid master key")
+	}
+
+	return &Slip77{
+		MasterKey: masterKey,
+	}, nil
+}
+
+// FromSeed derives the master key from the given seed and uses it to create
+// and return a new Slip77 instance
+func FromSeed(seed []byte) (*Slip77, error) {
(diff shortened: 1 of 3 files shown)


slip77/slip77.go
```diff
@@ -0,0 +1,69 @@
...
+	hmacRoot := hmac.New(sha512.New, domain)
+	hmacRoot.Write(seed)
+	root := hmacRoot.Sum(nil)
+
+	hmacMasterKey := hmac.New(sha512.New, root[:32])
+	hmacMasterKey.Write([]byte{prefix})
+	hmacMasterKey.Write(label)
+	masterKey := hmacMasterKey.Sum(nil)
+
+	return FromMasterKey(masterKey[32:])
```
tiero, line 49: there is no upper bound, better to slice to exactly 64th?

altafan: the `Sum` method always return a 64 byte array in this case

PR #102, opened 2020-06-11
Fix confidential.RangeProof method 

### LINKED ISSUE #96: [Confidential] Mandatory change to RangeProof
While studying how to blind an issuance transaction, I discovered that, when producing the range proof for blinding the asset/token amount of an input issuance, only one (private) key is used, rather than as done for the outputs where an ephemeral private key and a public key are used.
[This](https://github.com/ElementsProject/elements/blob/master/src/blind.cpp#L464) is taken from the elementsd implementation, while [this](https://github.com/ElementsProject/elements/blob/master/src/blind.cpp#L540) shows how the nonce is generated right after, when generating the proofs for the outputs.

We taylored our `RangeProof` function of the *confidential* package to always expect a private and a public key, for then generating the ecdh hash passed as a [*nonce*](https://github.com/vulpemventures/go-elements/blob/master/confidential/confidential.go#L178) to the CGO binding function.

We basically discovered that this so called *nonce* can be an ecdh hash of a key pair, or just a private key, and we support only the first option at the moment.

We could change directly the implementation of our `RangeProof` function by checking if the provided public key is nil, and in that case setting the nonce to the provided private key, otherwise, it would work as it does right now.

@tiero @sekulicd do you have other ideas?

altafan:
This changes the behavior of the `RangeProof` method of the confidential package so that it expects a nonce as argument of the function instead of creating fomr a couple of private and public keys.

This closes #96.

Please @tiero, review this.

### DIFF
--- a/confidential/confidential_test.go
+++ b/confidential/confidential_test.go
@@ -4,12 +4,13 @@ import (
 	"crypto/rand"
 	"encoding/hex"
 	"encoding/json"
-	"github.com/stretchr/testify/assert"
-	"github.com/vulpemventures/go-secp256k1-zkp"
 	"io/ioutil"
 	"math/big"
 	"strconv"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/vulpemventures/go-secp256k1-zkp"
 )
 
 var tests map[string]interface{}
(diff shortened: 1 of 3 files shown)


confidential/confidential_test.go
```diff
@@ -4,12 +4,13 @@ import (
...
 	"encoding/json"
-	"github.com/stretchr/testify/assert"
-	"github.com/vulpemventures/go-secp256k1-zkp"
 	"io/ioutil"
 	"math/big"
 	"strconv"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/vulpemventures/go-secp256k1-zkp"
```
tiero, line 13: why you need to import this?

altafan: they were already imported, I just reordered the imports so that we kee p builtin packages and "external" ones separated.

Anyway, assert is used to check that results match the expected values, while go-secp256k1-zkp is used to parse pedersen commitments from string vectors to *Commitment instances.

PR #108, opened 2020-06-17
Pset blinding transaction issuance inputs

### LINKED ISSUE #97: [PSET] Blinding transaction issuance inputs
The blinder role at the moment is represented by a `BlindOutputs` function that, as the name suggests, takes care of blinding **only the outputs** of a transaction.

When blinding a transaction that contains one or more issuances, instead, the blinder should blind **also the inputs** of the transaction that contain an issuance.

The issuance block contained by a tx input defines the asset total amount issued and the number reissuances that can be done, along with other information. These 2 amounts need to be substituted by commitments when blinding the transaction and also the respective range proofs must be added. 

So, blinding an input issuance means producing a commitment and a range proof for the asset amount and also for the token amount (if > 0) just as done for the outputs except for the fact that here it is not needed to calculate a final vbf.
So, at least for the asset amount, the blinder should do the following steps:
1. generate a random value blinding factor
2. generate a random asset blinding factor
3. generate the asset commitment (needed only for the following step)
4. generate the value commitment
5. generate the range proof

The last step deserves a focus:
as explained in #96, we don't need to calculate an ecdh hash from a key pair when generating the range proof here, but rather only a private key.
This means that, along with the changes that will be made to fix the linked issue, the `blinder` type needs to be extended with a new field `issuanceBlindingPrivKeys [][2][]byte`.
This weird type is due to the fact that the proofs for asset amount commitment and the token amount commitment (if > 0) should be generated by using 2 different keys. The public keys from them should be used to blind the issuance outputs of the transaction.

### LINKED ISSUE #105: Add blindInputs function to blinder
Taking into account all the considerations of #97 we should add to the `*blinder` type one unexported method `blindInputs` that takes care of blinding all the issuances contained in the inputs of the transaction (1 input can have only 1 issuance, remember).

The fields that must be blinded in a input issuance are the asset and the token amount.
It's worth to remember that an issuance can have a not defined token amount, thus, in this case, only the asset amount has to be blinded, while a reissuance has only an asset amount that needs to be blinded.

For blinding an asset amount it's necessary to do the following steps:
1. generate a random value blinding factor
1. generate a random asset blinding factor
1. generate the asset commitment (needed only for the following step)
1. generate the value commitment
1. generate the range proof

The same steps needs to be repeated, when necessary, for blinding also a token amount.

Now it's important to note that here for generating the range proof it's required a blinding private key, that will be used as the `nonce` to pass to the `confidential.RangeProof` method.
When blinding an output, instead, we use a blinding pubkey and an ephemeral private key to produce a ecdh nonce hash. This just to make evident the difference between the generation of the 2 range proofs.

Said all these things, the blinder role should be updated by adding one more field:
```
issuanceBlindingPrivKeys [][2][]byte
```
this represents a list of couples of private blinding keys, one couple for each input containing an issuance.
A couple of private keys because one is used for blinding the asset amount, and one to do the same the token amount (when this is not defined this key will simply be nil, or an empty slice).

Finally, the signature of the new method should be something like:
```
func (*blinder) blindInputs() error 
```

### LINKED ISSUE #106: Rename BlindOutputs into Blind
Taking into account what has been said in #97 we need to rename the actual `BlindOutputs` method of the `*blinder` type into `Blind`.

This because the method in question must be updated so that it calls the `blindInputs` function if any issuance blinding private key has been specified for any input.

It is important to note that the blinding of input issuances or reissuances must depend on the private keys passed, rather than making it mandatory when blinding a transaction because our goal should be letting the user be free of deciding what to blind in a transaction.

sekulicd:
Before this PSET blinder was only able to blind outputs.

After method blinder.BlindTransaction() is added it is possible to blind issuance in inputs and outputs.

Note that refactoring in issuance.go were done in order to be able to calculate asset/token hashes using new factory method NewTxIssuanceFromContractHash and invoking:
GenerateEntropy, GenerateAsset/GenerateReissuanceToken in sequence.

It closes #97 
It closes #105 
It closes #106 


sekulicd:
Currently TestCreateBlindIssuanceAndOutputSignAndBroadcast is not working.
There are no issues with blinding inputs, however, blinding outputs are failing(secp256k1.SurjectionProofInitialize) for the asset, which is added in AddIssuance method.

pset/blinder.go
```diff
@@ -113,6 +136,178 @@ func (b *blinder) BlindOutputs() error {
...
+		Exp:                 0,
+		MinBits:             52,
+	}
+	rangeProof, err := confidential.RangeProof(rangeProofInput)
+	if err != nil {
+		return err
+	}
+
+	b.pset.UnsignedTx.Inputs[index].
+		Issuance.AssetAmount = assetCommitment[:]
```
altafan, line 237: this should be set to the `valueCommitment`:
```suggestion
		Issuance.AssetAmount = valueCommitment[:]
```

pset/blinder.go
```diff
@@ -113,6 +136,178 @@ func (b *blinder) BlindOutputs() error {
 	return nil
 }
 
+//BlindOutputs method blinds pset's issuance pseudo input
+func (b *blinder) blindInputs() error {
+	for index, input := range b.pset.UnsignedTx.Inputs {
+		if input.Issuance != nil {
+			if b.issuanceBlindingPrivateKeys == nil {
+				return nil
+			}
```
altafan, line 145: this could be done before iterating over the inputs

pset/blinder.go
```diff
@@ -113,6 +136,178 @@ func (b *blinder) BlindOutputs() error {
...
+	)
+	if err != nil {
+		return err
+	}
+
+	abf, err := generateRandomNumber()
+	if err != nil {
+		return err
+	}
+	tokenAmount := input.Issuance.TokenAmount
```
altafan, line 258: it would be better to make a copy of `input.Issuance.TokenAmount`

pset/blinder.go
sekulicd: Done

pset/blinder.go
sekulicd: Done

pset/blinder.go
sekulicd: Done

altafan:
The problem with this PR was with the generation and usage of the blinding factors for the asset and token amount of an input issuance.

The first bug was that we were generating a random asset blinding factor, while this is required to be 0 since only the amounts of issuance are blinded.

The second problem was that we were not using the issuance amounts and their respective blinding factors for producing then blinding the outputs.

As explained [here](https://elementsproject.org/features/issued-assets/investigation), the output's surjection proof is needed to verify that even if all blinded, the asset of the inputs match all the assets of the outputs.
Since in an issuance tx there are up to 2 extra outputs, asset out and token out, we need to use the same number of pseudo inputs created from the issuance input for generating the correct output surjection proofs.

PR #119, opened 2020-07-01
Minor fixes to confidential, internal and pset packages

### LINKED ISSUE #104: [PSET] In Updater Upsbt should be renamed to Data
This is not critical, but should be better renaming to `Data` since it's exported and will be part of the public API. Could be misleading for the users.

### LINKED ISSUE #118: [Pset] fix blinder to blind all outs of the partial transaction
The behavior of the blinder should be changed in order to blind all the outputs in the partial transaction.
At the moment, instead, it expects that one more extra fee output is already present in the partial transaction, leaving it unblinded of course.

If we make this change we implicitly improve the flow of the pset package because the user would create his transaction with ins/outs to blind, then after the blinder have done its job, he could simply add all the unblinded outputs he wants, fees included (mandatory by the protocol, otherwise he wouldn't be able to broadcast the transaction).

altafan:
* This renames some of the fields of the `UnblindOutputArg` type of the confidential package. The `ValueCommit` and `Asset` became `ValueCommitment` and `AssetCommitment` so that it's more clear what this type expects.  
Also, the `UnblindOutput` function has been updated by converting the provided value commitment into a secp256k1 commitment. Prios this, the user was responsible for this conversion before calling the function, so user-friendliness has been improved.
* The internal `ReverseBytes` function has been fixed by returning a reversed copy of the provided byte slice instead of directly altering the one provided.
* The pset's updater field `Upsbt` has been renamed `Data`.  
The pset's blinder has been updated so that it blinds all the outputs of the tx. Unblinded ones can simply be added after the blinding.

This closes #104.
This closes #118.

Please @tiero, review this.


pset/blinder.go
```diff
@@ -100,7 +100,7 @@ func (b *blinder) validate() error {
 		)
 	}
 
-	if len(b.blindingPubkeys) != (len(b.pset.Outputs) - 1) {
+	if len(b.blindingPubkeys) != len(b.pset.Outputs) {
```
tiero, line 103: So this way now, if someone passes the fee output BEFORE blinding, this will not pass the validation. Wouldnt be better to find the fee output and exclude it from the lenght comparisons if found?

altafan:
I don't know why the fee output should be treated in a different way respect the other unblinded outputs of a transaction.
The updater's responsibility, generally speaking, is to correctly alter the partial transaction to add more info to it. Considering fee output present or not in the transaction that has to be blinded should just be an implementation decision, not something that I expect being defined in the spec (if ever one will be published 😅).

pset/blinder.go
altafan: We force the user to follow the following flow:
1. create the transaction with all inputs/outputs that must be blinded
2. blind the transaction
3. add all outputs that must be left unblinded

> Wouldnt be better to find the fee output and exclude it from the lenght comparisons if found?

if we support this eventuality, how do we know if the user has added also other outputs that he wants to keep unblinded?
We would open the door to other possible cases that would end up increase the complexity of the code with the purpose of just letting the user make more mistakes when creating/blinding his transaction.

I do believe that the above flow is the best option for the user, more than for our development, because it marks a clear separation between the blind and the unblind parts of a transaction.

tiero: Fee output = empty bytes. My point was that before this change "potentially" other apps had to add the fee output before and it means now breaking change. 
Practically no-one its using the library yet, so if we down this route "may" be OK

altafan: Yes, I think this is the right moment to do such this change cause the lib is still unknown basically.

sekulicd: From my POV this solution is acceptable, but I think that then we would need to give a better explanation of [this](https://github.com/altafan/go-elements/blob/6af125c9ff445d26d682966c5d1211ec4f2308f5/pset/blinder.go#L105) error message, maybe something like below:
"_blinding public keys do not match the number of outputs, note that fee and outputs that are not meant to be blinded should be added after blinder.Blind()_"
Maybe even more detailed message/

There is also option to enhance Updater method:
From:
`(p *Updater) AddOutput(txOutput *transaction.TxOutput)
`To:
`(p *Updater) AddOutput(txOutput *transaction.TxOutput, blind bool)
`
and then to add a new field, 
`outputsToBlind map[int] bool 
`
inside Updater and Blinder struct which would keep index of output and indicator if the output is supposed to be blinded.
This way we would have info which output user wants to blind and based on it to do filtering where it is needed.
@tiero @altafan

altafan: > From my POV this solution is acceptable, but I think that then we would need to give a better explanation of this error message, maybe something like below:

This is something that definitely could help the user understanding, thanks dusan.

> (p *Updater) AddOutput(txOutput *transaction.TxOutput, blind bool)

This is something that would be doable indeed, but when I say:
>We would open the door to other possible cases that would end up increase the complexity of the code with the purpose of just letting the user make more mistakes when creating/blinding his transaction.

I'm talking exactly about this: we could indeed increase the complexity of our updater/blinder, but the result will always be letting the user to mix "unblinded" and "blinded" outputs with the only disadvantage of increasing the risk that the user will experience problems with the blinding. No advantages at all IMO on having outputs that must be ignored when blinding the transaction, while they can safely be added in a second moment.

tiero: Agreed.

tiero: Could you update the error message as Dusan requested? Thanks

altafan: done.

PR #121, opened 2020-07-07
Issuance: Apply user given precision

### LINKED ISSUE #120: Fix usage of precision for issuance amount
At the moment the usage of the precision specified in the `NewTxIssuance` factory function is wrong in the sense that it should be used as a factor rather than divider in the formula. This should be:
```
assetAmount := arg.AssetAmount * 10^(p) # where p = precision = [0,8]
```

altafan:
This closes #120.

@tiero please, review this

### DIFF
--- a/transaction/issuance.go
+++ b/transaction/issuance.go
@@ -171,7 +171,7 @@ func (issuance *TxIssuanceExtended) GenerateReissuanceToken(flag uint) ([]byte, 
 }
 
 func toConfidentialAssetAmount(assetAmount float64, precision uint) ([]byte, error) {
-	amount := uint64(assetAmount * math.Pow10(int(8-precision)))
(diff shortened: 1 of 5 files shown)


transaction/issuance.go
```diff
@@ -170,21 +170,21 @@ func (issuance *TxIssuanceExtended) GenerateReissuanceToken(flag uint) ([]byte,
 	return token[:], nil
 }
 
-func toConfidentialAssetAmount(assetAmount float64, precision uint) ([]byte, error) {
-	amount := uint64(assetAmount * math.Pow10(int(8-precision)))
+func toConfidentialAssetAmount(assetAmount uint64, precision uint) ([]byte, error) {
+	amount := assetAmount * uint64(math.Pow10(int(8+precision)))
```
tiero, line 174: Just Pow10(precision)

PR #130, opened 2020-08-31
Pset validate signatures

### LINKED ISSUE #129: [PSET] validateSignaturesOfInput and validateSignaturesOfAllInputs
Would be very useful to have a method in the PSET type that checks if a signature is valid for a given input and also another one to check all inputs in the transaction. 

Useful when some user is expecting a pset encoded by someone else and he only need to finalize/extract the final tx.

Ref: https://github.com/provable-things/liquidjs-lib/blob/master/ts_src/psbt.ts#L322

sekulicd:
This adds two public methods on Pset struct:
ValidateAllSignatures() -> verifies all inputs signatures
ValidateInputSignatures() -> veifies only one input signature

Please review [TODO](https://github.com/sekulicd/go-elements/blob/d36da36b5d9dcf470ed0fbe3cf8d9b4977586036/pset/pset.go#L470) check I left for verifyScripForPubKey method.

It closes #129 


pset/pset.go
```diff
@@ -283,6 +289,200 @@ func (p *Pset) SanityCheck() error {
...
+
+	input := p.Inputs[inputIndex]
+
+	if input.NonWitnessUtxo != nil {
+		prevoutHash := p.UnsignedTx.Inputs[inputIndex].Hash
+		utxoHash := input.NonWitnessUtxo.WitnessHash()
+
+		if bytes.Compare(prevoutHash, utxoHash.CloneBytes()) == 1 {
+			return nil, nil,
+				errors.New("mon-witness utxo hash for input doesnt match the " +
```
tiero, line 377: typo: non-wintess

payment/payment.go
```diff
@@ -146,22 +146,21 @@ func FromScript(
...
-	switch script[0] {
-	// p2wpkh / p2wsh
-	case txscript.OP_0:
-		// in case the script is a p2wpkh, we need to set also the legacy script
+	switch address.GetScriptType(script) {
+	case address.P2WpkhScript:
 		if len(script[2:]) == 20 {
 			scriptHash = append(scriptHash, script[2:]...)
 			_script = buildScript(scriptHash, "p2pkh")
 		}
```
altafan, line 154: This statement checks that the script is of type p2wpkh actually.
With the new changes, this if can be deleted since we are sure the script is a p2wpkh in this case section.
```suggestion
		scriptHash = append(scriptHash, script[2:]...)
		_script = buildScript(scriptHash, "p2pkh")
```

pset/pset.go
```diff
@@ -283,6 +289,200 @@ func (p *Pset) SanityCheck() error {
...
+				input.WitnessScript,
+				prevout.Value,
+				txscript.SigHashType(sigHashType),
+			)
+			script = input.WitnessScript
+
+		case address.P2WpkhScript:
+			pay, err := payment.FromScript(
+				script,
+				&network.Liquid,
```
tiero, line 403: Is this FromScript used as a "validation" check for the given script? this script == pay.Script (at line 411?) 
Because otherwise we need to manage the network, provably inherit it from Pset struct (which is missing a network field)

pset/pset.go
```diff
@@ -283,6 +289,200 @@ func (p *Pset) SanityCheck() error {
...
+			script = input.RedeemScript
+		} else {
+			script = input.WitnessUtxo.Script
+		}
+		switch address.GetScriptType(script) {
+
+		case address.P2WpkhScript:
+			pay, err := payment.FromScript(
+				script,
+				&network.Liquid,
```
tiero, line 437: Same as before.

pset/pset.go
```diff
@@ -283,6 +289,200 @@ func (p *Pset) SanityCheck() error {
...
+		}
+		if !valid {
+			return false, nil
+		}
+	}
+	return true, nil
+}
+
+func (p *Pset) ValidateInputSignatures(inputIndex int) (bool, error) {
+	for _, partialSig := range p.Inputs[inputIndex].PartialSigs {
```
altafan, line 306: it should be better to first check that the input actually contains some signature, otherwise this method would return true for inputs that don't have any

pset/pset.go
```diff
@@ -283,6 +289,200 @@ func (p *Pset) SanityCheck() error {
...
+		return nil, nil, errors.New("need a utxo input item for signing")
+	}
+
+	return hash[:], script, nil
+}
+
+//TODO: check if this is valid only for P2PKH and P2WPKH since we
+//should not be able to compare script hash with public key hash
+func (p *Pset) verifyScripForPubKey(script []byte, pubKey []byte) (bool, error) {
+	pay, err := payment.FromScript(script, &network.Liquid, nil)
```
tiero, line 471: The network should be a parameter

pset/pset.go
```diff
@@ -283,6 +289,200 @@ func (p *Pset) SanityCheck() error {
...
+	} else {
+		return nil, nil, errors.New("need a utxo input item for signing")
+	}
+
+	return hash[:], script, nil
+}
+
+//TODO: check if this is valid only for P2PKH and P2WPKH since we
+//should not be able to compare script hash with public key hash
+func (p *Pset) verifyScripForPubKey(script []byte, pubKey []byte) (bool, error) {
```
tiero, line 470: typo: scripT

pset/pset.go
```diff
@@ -283,6 +289,200 @@ func (p *Pset) SanityCheck() error {
...
+		}
+
+	} else {
+		return nil, nil, errors.New("need a utxo input item for signing")
+	}
+
+	return hash[:], script, nil
+}
+
+//TODO: check if this is valid only for P2PKH and P2WPKH since we
```
tiero, line 468: cc @altafan

pset/pset.go
```diff
@@ -283,6 +289,200 @@ func (p *Pset) SanityCheck() error {
...
+		} else {
+			script = prevout.Script
+		}
+
+		switch address.GetScriptType(script) {
+
+		case address.P2WshScript:
+			hash = p.UnsignedTx.HashForWitnessV0(
+				inputIndex,
+				input.WitnessScript,
```
altafan, line 394: we must check that input.WitnessScript is not nil before using it

pset/pset.go
```diff
@@ -283,6 +289,200 @@ func (p *Pset) SanityCheck() error {
...
+			)
+			if err != nil {
+				return nil, nil, err
+			}
+			hash = p.UnsignedTx.HashForWitnessV0(
+				inputIndex,
+				pay.Script,
+				input.WitnessUtxo.Value,
+				txscript.SigHashType(sigHashType),
+			)
```
altafan, line 414: this parsing is useless, why not using simply `script`?
```suggestion
			hash = p.UnsignedTx.HashForWitnessV0(
				inputIndex,
				script,
				input.WitnessUtxo.Value,
				txscript.SigHashType(sigHashType),
			)
```

pset/pset.go
altafan: network here is useless here because we don't want an address but just the script, so it's just fine to pass it hardcoded

pset/pset.go
altafan: network is useless here because we don't want an address but just the script, so it's just fine to pass it hardcoded

altafan: this step actually creates the signing legacy pkh script `OP_DUP OP_HASH160 hash160(pk) OP_EQUALVERIFY OP_CHECKSIG` from the segwit version script `00 hash160(pk)`

pset/pset.go
```diff
@@ -283,6 +289,200 @@ func (p *Pset) SanityCheck() error {
...
+//TODO: check if this is valid only for P2PKH and P2WPKH since we
+//should not be able to compare script hash with public key hash
+func (p *Pset) verifyScripForPubKey(script []byte, pubKey []byte) (bool, error) {
+	pay, err := payment.FromScript(script, &network.Liquid, nil)
+	if err != nil {
+		return false, err
+	}
+
+	pkh := payment.Hash160(pubKey)
+	res := bytes.Compare(pay.Hash, pkh)
```
altafan, line 477: This is valid only for p2pkh/p2wpkh indeed.
In the reference implementation of [liquidJS](https://github.com/vulpemventures/liquidjs-lib/blob/master/ts_src/psbt.ts#L1124) the script is decompiled and then we check that the pkHash or the pk appears in the script. This because the script can contain:
* a full pubkey in case it's a P2PK or P2MS (in this case the script that arrives here is the redeem script containing all the pubkeys)
* a pkHash in case it's a P2PKH or a P2WPKH
* either a pkHash or a full pubkey for complex P2SH or P2WSH since for this kind of types, the script arriving here will be a redeem script (just like how happens for P2MS)


We should check if in btcutil / btcd repos there's something similar to `bScript.decompile()` that decompiles a a script into a sequence of opcodes/hashes

pset/pset.go
sekulicd: Done

payment/payment.go
sekulicd: Done

pset/pset.go
sekulicd: Done

pset/pset.go
sekulicd: Done

pset/pset.go
sekulicd: Done

pset/pset.go
sekulicd: Done

pset/pset.go
sekulicd: Done

pset/pset.go
sekulicd: Done

pset/pset.go
sekulicd: Done

pset/pset.go
```diff
@@ -289,9 +289,9 @@ func (p *Pset) SanityCheck() error {
 	return nil
 }
 
-func (p *Pset) ValidateAllSignatures() (bool, error) {
+func (p *Pset) ValidateAllSignatures(net *network.Network) (bool, error) {
```
altafan, line 292: Please revert these changes.
Network has nothing to do with signature verification. As I said in the previous comment, it's totally fine to hardcode either Liquid or Regtest because we are not interested in deriving addresses but just scripts/hashes. Network is useful just to prepend the correct address prefix when encoding the script into base58/bech32/blech32 format.

tiero: Agreed, better to pass nil, to avoid confusion.

pset/pset.go
altafan: I found this method [`txscript.DisasmString`](https://godoc.org/github.com/btcsuite/btcd/txscript#DisasmString) in btcd that parses a raw bytes script into a string of space separated op codes and data.
We could use this method to parse the script and then search for the pubkey of interest like:
```
scriptAsm, err = txscript.DisasmString(script)

if strings.Contains(scriptAsm, pubKey.SerializeCompressed() || strings.Contains(scriptAsm, pkHash) {
   ...
}
```

cc @tiero @sekulicd

sekulicd: Done

pset/pset.go
sekulicd: Done

altafan (approved):

PR #141, opened 2020-11-13
Verify surjection proof

altafan:
This adds the verification of the surjection proof once generated to make sure if it's valid.

If not, the blinder has been changed so that it can stop the blinding of the output without committing any change to the UnsignedTx fields of the partial transaction.

Tests of confidential transactions in pset/pset_test.go have been updated by retrying to blind the tx if necessary.

BONUS: made the BlindedSwap test a full confidential swap (input were unconfidential before this).

Please @tiero review this.

### DIFF
--- a/confidential/confidential.go
+++ b/confidential/confidential.go
@@ -238,7 +238,7 @@ type SurjectionProofArg struct {
 }
 
 //SurjectionProof method generates surjection proof
-func SurjectionProof(input SurjectionProofArg) ([]byte, error) {
+func SurjectionProof(input SurjectionProofArg) ([]byte, bool) {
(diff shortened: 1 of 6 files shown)


confidential/confidential.go
```diff
@@ -305,12 +305,22 @@ func SurjectionProof(input SurjectionProofArg) ([]byte, error) {
...
+		ctx,
+		proof,
+		inputGenerators,
+		*outputGenerator,
+	) {
+		return nil, false
 	}
 
-	return secp256k1.SurjectionProofSerialize(ctx, proof)
+	return proof.Bytes(), true
```
tiero, line 323: the method is changed?

tiero (approved):

PR #143, opened 2020-11-17
Verify proof method + Refactor confidential pkg

### LINKED ISSUE #142: Methods for external verification of surjection proofs
At the moment, the library does verify surjection proofs once generated, but doesn't let the user perform such verification.

We need to expose a `VerifySurjectionProof` method in the `confidential` pkg accepting in/out assets and asset blinders as bytes, a proof as byte that basically wraps the respective method of the go-secp256 library.

Also, a `ValidateBlinding` function needs to be added to the `pset`'s blinder expecting a pset object, and a list of in/out private blinding keys that unblinds all the inputs and outputs of the tx and calls the new method defined in the confidential pkg. This way we validate both range and surjection proofs at once.

altafan:
This adds methods to oth confidential and pset packages that allow the end-user to verify surjection proofs or an entire blinded transactions. Closes #142.

BONUS: this refactors the confidential package that was looking pretty obsolete so far. The existing methods haven't been mofidified but UnblindOutput which have been replaced by UnblindOutputWithKey and UnblindOutputWIthNonce.
Another method has been added along with `VerifySurjectionProof`, named `UnblindIssuance` (pretty self-expressive the purpose).
The functions helping to convert amounts from satoshi to elements format and vice-versa have been moved to a dedicated package to prevent import cycles between confidential and transaction packages.

Please @tiero, review this.


tiero:
@altafan UnblindOutput method has been used in many repos. This would break them/force to update.

Could we keep the same API/signature if possibile?

altafan:
I don't think it's possible since I removed the UnblindOutputArg type in favour of a simpler transaction.TxOutput + blindKey/ecdhNonce. We will update the codebase of the projects that already use go-elements.

pset/blinder.go
```diff
@@ -658,6 +642,123 @@ func (b *blinder) blindToken(index int, token, vbf, abf []byte) error {
 	return nil
 }
 
+func verifyBlinding(
```
sekulicd, line 645: is it better to return error instead bool? we can propagate error so client has more insight on why blinding was not valid

altafan: I think a bool is just enough. A transaction that is correctly blinded may be not valid in the sense that the proofs could be invalid, but it should be at least well-formed. This is not intended to be used to check that the blinding has been done correctly, but just to verify that the proofs are all ok.

tiero: Problem is that the error would not be meaningful for the user. If wrong, the only solution is to retry. The blinding is either correct or not.

tiero (approved):

PR #151, opened 2021-02-05
gomobile bind: Allow go-elements to be built for Android and iOS targets

tiero:
This commit allows to build `go-elements` packages as a single standalone library to be imported in iOS and Android targets using `gomobile bind`

To do so some cosmetic change is needed to make it work, most importantly we have Only a subset of Go types wich are currently supported and return variables should be at maximum 2 where the second shoudl alwasy be the `error`

The only breaking change has been for the `slip77` where `DeriveKey` now it returns a struct `KeyPair` instead of two variables. 

I see this as an acceptable tradeoff overall, instead of creating dedicated slip77 gomobile binding.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,6 @@
 
 # WebStorm IDE project config files
 .idea
+
+*.aar
+*.framework
--- a/.gitignore
+++ b/.gitignore
@@ -18,5 +18,7 @@
 # WebStorm IDE project config files
 .idea
 
+# gomobile artifacts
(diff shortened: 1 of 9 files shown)


tiero:
Unfortunately is not that easy, gomobile seems to not bind the entire library (ie. Payment, Pset etc..)

I think the best way is to creata a more high level library that uses go-elements and expose only the supported go types for `gomobile` Moving to `github.com/vulpemventures/ionio`

PR #159, opened 2021-04-02
Block pkg

### LINKED ISSUE #156: Add Block package to handle (de)serialization of raw Elements blocks
What we need is a way to de-serialize a block received from network, in order to create a list of go-elments's `Transaction`

Data structures 

https://github.com/btcsuite/btcd/blob/master/wire/blockheader.go#L22
https://github.com/btcsuite/btcd/blob/master/wire/msgblock.go#L64

An implementation for Liquid could be found in Elements C codebase, but I find Rust more enjoyable to read
https://github.com/ElementsProject/rust-elements/blob/master/src/block.rs

sekulicd:
This adds block package with option to serialize and deserialise raw and hex block representation to Block struct.

This  closes #156


tiero:
I would say lets test the deserialization first.

sekulicd:
> I would say lets test the deserialization first.

@tiero I added some tests, check if this is ok.
serialization also added

block/deserialize.go
```diff
@@ -0,0 +1,341 @@
...
+
+	version, err := d.ReadUint32()
+	if err != nil {
+		return nil, err
+	}
+
+	isDyna := false
+	if version>>31 == 1 {
+		isDyna = true
+	}
```
altafan, line 44: this could become: 
```suggestion
	isDyna := version>>31 == 1
```
couldn't it?

block/deserialize.go
```diff
@@ -0,0 +1,341 @@
...
+	version = blockVersion
```
altafan, line 45: can you explain why here the deserialized version is overwritten by the constant one?

block/deserialize.go
```diff
@@ -0,0 +1,341 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	extData, err := deserializeExtData(d, isDyna)
+	if err != nil {
+		return nil, err
+	}
+
+	extData.IsDyna = isDyna
```
altafan, line 72: couldn't this be done into deserializeExtData? Like:
```
return &ExtData{
	Proof:             proof,
	DynamicFederation: dynamicFederation,
        IsDyna: isDyna,
}
```

block/deserialize.go
```diff
@@ -0,0 +1,341 @@
...
+	nul := false
+
+	serializeType, err := d.ReadUint8()
+	if err != nil {
+		return nil, err
+	}
+
+	switch serializeType {
+	case null:
+		nul = true
```
altafan, line 150: why not directly `return nil, nil` here so we can get rid of this `nul` variable? If I didn't get it wrong, in case serializeType is null, dynamicFederationParams is nil and it's returned just like that, isn't it?

block/deserialize.go
```diff
@@ -0,0 +1,341 @@
...
+import (
+	"bytes"
+	"errors"
+
+	"github.com/vulpemventures/go-elements/internal/bufferutil"
+
+	"github.com/vulpemventures/go-elements/transaction"
+)
+
+func deserialize(buf *bytes.Buffer) (*Block, error) {
```
altafan, line 12: why this function is unexposed?

block/deserialize_test.go
```diff
@@ -0,0 +1,173 @@
...
+			assert.Equal(
+				t,
+				v["numOfTx"].(string),
+				strconv.Itoa(len(block.TransactionsData.Transactions)),
+			)
+		})
+	}
+}
+
+func TestBlockDeserializationIntegration(t *testing.T) {
```
altafan, line 50: I think this might be overkill.
I mean, IMO it's enough to have fixtures with dynafed/non-dynafed blocks and make sure that they are correctly deserialized. I don't see any added value in creating a tx, broadcast it and deserialize the raw block fetched from some explorer. This just adds some (most likely a lot) overhead to the test, but basically checks the same thing of `TestDeserialization`.

block/serialize.go
```diff
@@ -0,0 +1,301 @@
...
+func (b *Block) SerializeBlock() ([]byte, error) {
+	s, err := bufferutil.NewSerializer(nil)
+	if err != nil {
+		return nil, err
+	}
+
+	err = b.Header.SerializeHeader(s)
+	if err != nil {
+		return nil, err
+	}
```
altafan, line 16: can you change to this code-style here but also elsewhere in this package:
```suggestion
	if err := b.Header.SerializeHeader(s); err != nil {
		return nil, err
	}
```

block/serialize.go
```diff
@@ -0,0 +1,301 @@
...
+
+	err = h.ExtData.serialize(s)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (e *ExtData) serialize(
```
altafan, line 91: I think it would be better to make the ExtData methods public.

block/serialize_test.go
```diff
@@ -0,0 +1,78 @@
...
+			if err != nil {
+				t.Error(err)
+			}
+
+			assert.Equal(t, v["hex"].(string), hex.EncodeToString(serializeBlock))
+		})
+	}
+}
+
+func TestBlockHeaderSerialization(t *testing.T) {
```
altafan, line 41: The previous test should already test the header serialization, shouldn't it?
Is this test necessary?

block/testdata/deserialize.json
```diff
@@ -0,0 +1,23 @@
+[
```
altafan, line 1: What's the origin of these fixtures? Did you already check if we can reuse some from the elements repository? It's likey that they already have fixtures for these operations and would be cool to share the same test vectors.

altafan (changes requested):

block/deserialize_test.go
sekulicd: I just wanted to make sure that beside header also transactions are deserialised correctly.

block/testdata/deserialize.json
sekulicd: I used some of fixtures from [Rust repo](https://github.com/ElementsProject/rust-elements/blob/0d67c57afa1137ab27861bb8c2190413929d4301/src/block.rs#L383) also added one from local environment.

block/deserialize.go
sekulicd: In [Rust impl](https://github.com/ElementsProject/rust-elements/blob/0d67c57afa1137ab27861bb8c2190413929d4301/src/block.rs#L315) version is overwritten in case of isDyna true but even if it is false value is equal to blockVersion, anyways to be on the same page i will do the same.

block/deserialize.go
sekulicd: Done

block/deserialize.go
sekulicd: Done

block/deserialize.go
sekulicd: Done

block/deserialize.go
sekulicd: Since I have NewFromHex and NewFromBuffer, which is only invoking deserialise

block/serialize_test.go
sekulicd: You are right, this test case was helping me to identify mistakes easier since transaction was not included.
I will remove it.

block/serialize.go
sekulicd: @altafan  do u think there will be need to serialize something from ExtData separately from Header?

block/serialize.go
sekulicd: Done

block/deserialize.go
sekulicd: Done

block/serialize_test.go
sekulicd: Done

block/deserialize.go
altafan: Ah ok, but why other functions here have been made public then?

sekulicd: I made public DeserializeHeader and DeserializeTransactions, thought i might be useful  ...

block/deserialize_test.go
altafan: couldn't we use fixtures also for this? By the way, I thought this was already ensured by the test above

sekulicd: i wanted to check value if it is correctly deserialized, to have live integration test, in fixtures i used, i would need to have blinding keys which i dont, you think we should remove it?

block/serialize.go
altafan: I'm sorry, I was thinking ExtData was at the same level of Header and Transactions. It's fine as it is now. Resolving.

block/testdata/deserialize.json
altafan: cool! :) would be nice to add this reference somewhere in the code, like in one of the test file.

sekulicd: Done

block/deserialize.go
altafan: So, I guess the current `blockVersion`value is the result of some block version & 0x7fffffff.. Are we sure liquid blocks have always the same version though? If yes, I recommend adding a comment before this assignment to specify why it's done.
Otherwise, we should better follow rust-elements and make an AND bit-a-bit operation using the same hex value declared there.

What do you think?

sekulicd: Lets fellow Rust elements.
Done

PR #161, opened 2021-04-26
[WIP]Pegin

### LINKED ISSUE #158: Derive Pegin script from federation multisignature

sekulicd:
Pegin process consists of two steps:
1. Invoking getpeginaddress RPC - which returns mainchain_address (address in BTC network) to which BTC should be sent and claim_script which should be used with other params in claimpegin RPC
2. Invoking claimpegin RPC with with user 'claims' LBTC equals to the amount of BTC send in first request.

This commit contains implementation for getpeginaddress RPC.
It also contains btcd modified methods for parsing scripts.

This closes #158 


pegin/getaddress.go
```diff
@@ -0,0 +1,379 @@
...
+
+type AddressInfo struct {
+	ClaimScript      string
+	MainChainAddress string
+}
+
+// GetAddressInfo simulates getpeginaddress RPC call which returns mainChain
+// address to which BTC should be sent in BTC network and claim script that is
+// to be used in claimpegin RPC alongside with bitcoinTx and txoutproof
+func GetAddressInfo(
```
tiero, line 43: This is ok for testing, but most likely we will not expose this API "one-shot". Will eventually do in a separate repo/utility package if we need in Go posix environments

pegin/getaddress.go
```diff
@@ -0,0 +1,379 @@
...
+				ctx,
+				maybeTweaked2,
+				secp256k1.EcUncompressed,
+			)
+
+			if !bytes.Equal(tweaked2Bytes[:64], maybeTweaked2Bytes[:64]) {
+				return nil, errors.New("sanity check failed")
+			}
+
+		} else {
```
tiero, line 254: Whats' the rationale here? I know is monkey porting of C version, but anything as context you can add would be appreciate.

pegin/getaddress.go
```diff
@@ -0,0 +1,379 @@
...
+	FedpegScript  []byte
+	FedpegProgram []byte
+}
+
+// TODO - get fedpeg details from live blockchain
+func GetFedpegInfo() (*FedpegInfo, error) {
+	return &FedpegInfo{}, nil
+}
+
+func calculateContract(
```
tiero, line 158: this method seems the one maybe to be split, so other targets could use their own secp. Maybe we should decompose calculate contract in many little functions that only do data structure manipulation and expect data as input?

Like, generateContract(watchman, tweaked, ...) 


Any other idea?

sekulicd: As agreed, contract func extracted to other pkg

pegin/getaddress.go
sekulicd: Done

pegin/getaddress.go
sekulicd: Dont have e2e knowledge(pegin, pegout, crypto) in order to add relevant comments in here.

address/opcode.go
```diff
@@ -0,0 +1,567 @@
+package address
+
```
altafan, line 2: Aren't all these opcodes already importable from [btcd/txscript](https://github.com/btcsuite/btcd/blob/master/txscript/opcode.go)? Why do we need to redefine them here?

address/script.go
```diff
@@ -0,0 +1,350 @@
+package address
+
```
altafan, line 2: same for this file.

pegincontract/contract.go
```diff
@@ -0,0 +1,238 @@
...
+	federationScript []byte,
+	scriptPubKey []byte,
+) ([]byte, error) {
+	contract := make([]byte, 0)
+
+	ctx, err := secp256k1.ContextCreate(secp256k1.ContextBoth)
+	if err != nil {
+		return nil, err
+	}
+
```
altafan, line 27: add `defer ctx.Destroy()`

address/opcode.go
sekulicd: Yes, opcodeArray on the other hand is not exported and it is needed in ParseScript method which is also unexported method in btcd

address/script.go
sekulicd: ExtractScriptAddrs method was slightly modified ExtractPkScriptAddrs which at one point I needed(to check if fedpeg script is multisig etc) but later i saw that i dont need to use it, so now i remove it.

pegincontract/contract.go
sekulicd: Done

sekulicd:
> Why having 2 packages `pegin` and `pegincontract`? wouldn't it be better to merge them and just have `pegin`?

cause pegincontract is coupled to C libs which is problematic for WASM compliation.

pegin/getaddress.go
```diff
@@ -0,0 +1,142 @@
...
+
+	return liquidNetwork, btcNetwork
+}
+
+func getClaimWitnessScript(
+	privateKey btcec.PrivateKey,
+	net *network.Network,
+) string {
+	p2wpkh := payment.FromPublicKey(
+		privateKey.PubKey(),
```
tiero, line 85: no need to ask for private key secret material, just the public seems ok to me to generate the claiming script

pegin/getaddress.go
```diff
@@ -0,0 +1,142 @@
...
+type AddressInfo struct {
+	ClaimScript      string
+	MainChainAddress string
+}
+
+// GetAddressInfo simulates getpeginaddress RPC call which returns mainChain
+// address to which BTC should be sent in BTC network and claim script that is
+// to be used in claimpegin RPC alongside with bitcoinTx and txoutproof
+func GetAddressInfo(
+	privateKey btcec.PrivateKey,
```
tiero, line 37: better using a []byte and/or string/hex since is an high level method already.

pegin/getaddress.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/vulpemventures/go-elements/payment"
+
+	"github.com/btcsuite/btcd/chaincfg"
+
+	"github.com/vulpemventures/go-elements/address"
+
+	"github.com/btcsuite/btcd/btcec"
+)
+
+const (
```
tiero, line 21: woudl be more self describing using string instead of integers here?

tiero (changes requested):

pegin/getaddress.go
sekulicd: Done

pegin/getaddress.go
sekulicd: Done

pegin/getaddress.go
sekulicd: I think that variable name is already self explanatory and its exported to it can be passed as arg to func GetAddressInfo

pegin/getaddress.go
louisinger: @sekulicd I would rename the function `GetPeginAddress`, GetAddressInfo can be confusing as there is a `getaddressinfo` RPC ?

pegin/getaddress.go
```diff
@@ -0,0 +1,151 @@
...
+	ClaimScript      string
+	MainChainAddress string
+}
+
+// GetAddressInfo simulates getpeginaddress RPC call which returns mainChain
+// address to which BTC should be sent in BTC network and claim script that is
+// to be used in claimpegin RPC alongside with bitcoinTx and txoutproof
+func GetAddressInfo(
+	publicKey []byte,
+	fedpegInfo FedpegInfo,
```
louisinger, line 38: actually, we only need the `FedPegInfo.FedPegScript` so the federation's script ([]byte) could replace `FedpegInfo` parameter

tiero: I would remove GetAddressInfo entirerely actually, ClaimWitnessScript and PeginAddress are enough high level API

pegin/getaddress.go
sekulicd: u are right current name is not the happiest one, it can be misleading, i was not happy with GetPeginAddress  since it returns two things mainchain_address and claim_script so i thought to have more general name

tiero: I suggest to remove it altogether https://github.com/vulpemventures/go-elements/pull/161#discussion_r625215790

sekulicd: Done

tiero (changes requested):
Can you rename pegin/getaddress.go to just pegin.go?

PR #165, opened 2021-05-20
Claim pegin

### LINKED ISSUE #162: Add method to create a claim pegin transaction

sekulicd:
This adds Clam function to pegin pkg which is creating unsigned raw transaction.

Please take a look at [e2e test](https://github.com/sekulicd/go-elements/blob/a3e8f52ac45e4fd77c9262415d8c208e0fdc5515/pegincontract/pegin_e2e_test.go#L34).

This closes #162 


pegin/pegin.go
```diff
@@ -75,44 +76,294 @@ func PeginAddress(
...
+	parentGenesisBlockHash []byte,
+	fedpegScript []byte,
+	contract []byte,
+	btcTx []byte,
+	btcTxOutProof []byte,
+	claimScript []byte,
+	receiverScript []byte,
+	milisatsPerByte float64,
+) (*transaction.Transaction, error) {
+	claimTx := &transaction.Transaction{}
```
altafan, line 107:
```suggestion
	claimTx := &transaction.NewTx(2)
```

pegin/pegin.go
```diff
@@ -75,44 +76,294 @@ func PeginAddress(
...
+	txSize := claimTx.VirtualSize()
+	feeValue := uint64(float64(txSize) * milisatsPerByte)
+
+	finalValue := uint64(amount) - feeValue
+	finalValueElements, err := elementsutil.SatoshiToElementsValue(finalValue)
+	if err != nil {
+		return nil, err
+	}
+
+	feeValueElements, err := elementsutil.SatoshiToElementsValue(finalValue)
```
altafan, line 146:
```suggestion
	feeValueElements, err := elementsutil.SatoshiToElementsValue(feeValue)
```

pegincontract/contract_test.go
```diff
@@ -114,3 +140,213 @@ func TestCalculateContract(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	var lbtc = append(
+		[]byte{0x01},
+		elementsutil.ReverseBytes(peggedAssetBytes)...,
+	)
+
+	parentBlockHash := "0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206"
+	parentBlockHashBytes, err := hex.DecodeString(parentBlockHash)
```
altafan, line 214:
```suggestion
	parentBlockHashBytes, err := hex.DecodeString(bufferutil.ReverseBytes(parentBlockHash))
```

altafan (changes requested):

altafan:
@tiero [in this comment](https://github.com/vulpemventures/go-elements/pull/161#discussion_r625215790) you suggested to split the function in 2 distinct methods.. can you explain the reason?

Personally, I think that a single function `NewPeginAddress` that returns both the btc address and the liquid claim script would be better because the only other tool that provides such a feature is the `getpeginaddress` command of the element's CLI. Maybe better to be consistent with it.

altafan:
@dusan all things you are doing in TestClaimPegin by making use of nigiri are achievable by making HTTP requests to `localhost:3001` (nigiri-chopsticks).

tiero:
> @tiero [in this comment](https://github.com/vulpemventures/go-elements/pull/161#discussion_r625215790) you suggested to split the function in 2 distinct methods.. can you explain the reason?
> 
> Personally, I think that a single function `NewPeginAddress` that returns both the btc address and the liquid claim script would be better because the only other tool that provides such a feature is the `getpeginaddress` command of the element's CLI. Maybe better to be consistent with it.

`getpeginaddress` is command line/JSONRPC method, we don't have a comparison with other SDK/libraries, so no need to be consistent, because we are not building a CLI wallet in the first place.

I suggested to split beacuse in the end, the claim peing script, is something we can achieve already with the current API on the `payment` package in go and potentially done with other libraires in other targets (GDK on iOS/Android, LiquidJS in browesers)

block/merkle_block.go
```diff
@@ -0,0 +1,274 @@
...
+// merkle tree at the given depth-first height.
+func (m *MerkleBlock) calcTreeWidth(height uint32) uint32 {
+	return (m.PartialMerkleTree.TxTotalCount + (1 << height) - 1) >> height
+}
+
+func deserializePartialMerkleTree(
+	buf *bytes.Buffer,
+	mb wire.MsgMerkleBlock,
+) (*PartialMerkleTree, error) {
+	//d := bufferutil.NewDeserializer(buf)
```
tiero, line 188: Remove commented code if not needed

pegin/pegin.go
```diff
@@ -18,19 +24,12 @@ import (
...
-
-type NetworkType int
-
 type AddressInfo struct {
 	ClaimScript      string
 	MainChainAddress string
 }
 
-func PeginAddress(
+func MainChainAddress(
```
tiero, line 32: This will eventually break the wasm package module it depends on it. Any reason to change the API name?

pegin/pegin.go
```diff
@@ -42,7 +41,7 @@ func PeginAddress(
 	}
 
 	var mainChainAddress string
-	if isDynaFedEnabled || address.IsScriptHash(pops) {
+	if !isDynaFedEnabled || address.IsScriptHash(pops) { //TODO check IsScriptHash
```
tiero, line 44: regression?

pegin/pegin.go
sekulicd: Package name is already named and its not good practice that method contains pkg name.

block/merkle_block.go
sekulicd: Done

pegin/pegin.go
sekulicd: Should i still use PeginAddress?

tiero: I see the point, just that now we depend on it. But since we touch again to add the Claim method, its ok

pegin/pegin.go
tiero: @sekulicd

sekulicd: need to double check address.IsScriptHash against elements implementation.
!isDynaFedEnabled is ok.

sekulicd: Verifying IsScriptHash Done, this line should be ok.

PR #171, opened 2021-06-16
Issuance hash bug 170

### LINKED ISSUE #170: Issuance tx_id issue
Seems that we have issue with generating tx_id for issuance tx.
Example in [here](https://gist.github.com/altafan/58c4c986f0419e1027906ac3f50de6b2)

sekulicd:
This adds additional method IssuanceHash() which correctly calculates tx hash in case of Issuance Transaction.

This closes #170 

@tiero @tiero please review.


transaction/transaction.go
```diff
@@ -356,6 +357,23 @@ func (tx *Transaction) TxHash() chainhash.Hash {
...
+		if v.Issuance != nil {
+			hasIssuance = true
+		}
+	}
+	if !hasIssuance {
+		return [32]byte{}, errors.New("no input with issuance")
+	}
+
+	// Encode the transaction and calculate double sha256 on the result.
+	buf, _ := tx.serialize(nil, false, false, false)
```
tiero, line 373: can you explain  why all false?

sekulicd: While debugging I noticed that when serialising, [this](https://github.com/sekulicd/go-elements/blob/4e90ba16eaa7ba396562acb763f57c123a33f678/transaction/transaction.go#L731) is not reached, cause 'zeroFlag' was true in TxHash() method, after I changed this flag to false, txHash was as expected.
Maybe @altafan can explain better what this flag is doing and if he has other idea on how to solve this.

altafan:
@sekulic I cloned your branch and added a commit to fix the bug within the TxHash method. There's no need of defining a new method.
Instead of creating a new test, I added 2 new test vectors to the fixtures consumed by TestTxHash: one for an issuance transaction and another for a pegin one.

You should be able to cherry-pick the commit 8a0f8001c4e6a7e51730e2c8a0ce0b8cab8cf2f4, in alternative just copy-paste the entire transaction.go, transaction_test.go and data/tx_valid.json files of the transaction pkg.

sekulicd:
Done

altafan (approved):

PR #174, opened 2021-09-13
Coin type

sekulicd:
This adds HDCoinType field inside network struct to be used for derivation path.


tiero:
nACK. I tought we had already coin type for BIP44 derivation, but if is not present, thats not something should stay here IMHO, and delegated to wallet developers to define on top

PR #181, opened 2021-11-17
Pset V2 package

sekulicd:
This adds updater and blinder role.


pset/updater.go
```diff
@@ -36,7 +36,7 @@ type Updater struct {
 }
 
 // NewUpdater returns a new instance of Updater, if the passed Psbt struct is
-// in a valid form, else an error.
+// in a valid form, else an error.a
```
tiero, line 39:
```suggestion
// in a valid form, else an error.
```

psetv2/blinder.go
```diff
@@ -0,0 +1,561 @@
+package psetv2
```
tiero, line 1: This is a more of a topic of design, one of the main purpose during this implementation, was being able to compile pset package to wasm ie. having a way to detach all the CGO operation out of the package.

We may need to have a dedicated pkg for blinder role maybe?

sekulicd: only option is to move it to other pkg but then we would also need to export all fields of Pset(Globa, Input, Output)

tiero: what if we do like the signer role, where we expect via methods to pass stuff to be added ie. AddSignature?

tiero: but I think is fine to export those fields maybe, for easy access?

confidential/confidential.go
```diff
@@ -492,7 +516,7 @@ func inAssetGenerators(inAssets, inAssetBlinders [][]byte) ([]*secp256k1.Generat
 
 	inGenerators := make([]*secp256k1.Generator, 0, len(inAssets))
 	for i, inAsset := range inAssets {
-		gen, err := secp256k1.GeneratorGenerateBlinded(
+		gen, err := secp256k1.GeneratorGenerateBlinded( //TODO in elements repo generate_blinded is not used?
```
altafan, line 519: it's used indeed https://github.com/ElementsProject/elements/blob/d973e1105d0eacda1b31316c1a58aec4a73029b2/src/blind.cpp#L348.

PR #183, opened 2021-11-22
Pubkey issuance contract

### LINKED ISSUE #182: the hex-encoded public key of the issuer is missing IssuanceContract struct
[Here](https://github.com/vulpemventures/go-elements/blob/71dd075222b5880722fcdb922fee3150cd5f383e/transaction/issuance.go#L20) we should include the Issuer Pub Key as weel 

reference specification

https://docs.blockstream.com/liquid/developer-guide/proof-of-issuance.html#contract-json-fields

sekulicd:
This adds issuer_pubkey to the Issuance.

This close #182

@tiero please review.


sekulicd:
Moving conversation from [here](https://github.com/vulpemventures/go-elements/issues/182) to this PR.

@tiero please check [test](https://github.com/sekulicd/go-elements/blob/4a5feb423bdd9b9568d58a4355ddc6b0cd9121d4/transaction/issuance_test.go#L94) which is passing contract hash validation.

Still old test is failing, @altafan any idea why? where did u find these fixtures?

transaction/issuance.go
```diff
@@ -219,3 +223,16 @@ func toConfidentialTokenAmount(tokenAmount uint64) ([]byte, error) {
 	}
 	return confAmount[:], nil
 }
+
+func orderJsonKeysLexicographically(bytes []byte) ([]byte, error) {
```
tiero, line 227:
```suggestion
func orderJsonKeysLexographically(bytes []byte) ([]byte, error) {
```

sekulicd: Done

PR #185, opened 2021-11-29
`block` pkg: add Hash method 

louisinger:
This PR adds a `forHash` bool parameter to the SerializeHeader method and thus generate the serialized block header buffer according to our need (for "reading" or for "hash")

- it lets to return the block's ID using the `Header.GetHash` method.
- tested in hash_hash.go (with Liquid & Nigiri block)

@tiero @sekulicd please review


tiero:
Wouldn't be better, API wise, to have two separated methods, each with a specific purpose explicit in the name, instead of a boolean parameter

tiero (changes requested):

PR #188, opened 2022-01-28
Fix parsing witness script to Payment with FromScript()

### LINKED ISSUE #186: Get the WitnessPubKeyHash from a Payment
Hi, 

I was wondering something. 

I am making P2WPKH payments with an application, and from another one I try to read the blocks and retrieve any payment made by the other application. 

```
data := []byte(msg)
x := bytes.NewBuffer(data)
b, err := block.NewFromBuffer(x)
if err != nil {
	fmt.Println(err)
	return
}
fmt.Println("-- BLOCK")
for tx := range b.TransactionsData.Transactions {
	fmt.Println("---- TXN")
	for output := range tx.Outputs {
		fmt.Println("------ OUTPUT")

		if output.Script == nil || len(output.Script) == 0 {
			return
		}

		switch address.GetScriptType(output.Script) {
		case address.P2WpkhScript:
			fmt.Println("------ P2WPKH")

			p, err := payment.FromScript(output.Script, &network.Regtest, nil)
			if err != nil {
				fmt.Println(err)
				continue
			}

                        // same outputs
			fmt.Println(p.WitnessPubKeyHash())
			fmt.Println(p.WitnessScriptHash())

		default:
			fmt.Println("------ OTHER")
			continue
		}
	}
}
````

So my first question is why the output of the two lines: 
```
			fmt.Println(p.WitnessPubKeyHash())
			fmt.Println(p.WitnessScriptHash())
```
is the same? 

And my second question is that somehow the output is the same, but actually it's not what I am expecting. 
I get a long address of 75 bytes, while the payments are made on addresses of 43 bytes. So I am expecting an address of 43 bytes. 

Looking more carefully at the output, I noticed that the addresses are matching minus the "checksum":
````
Original address:      ert1q989xa8zz0a4fs7zrsmfsacwkvnup8vyd 27xnhv
p.WitnessPubKeyHash(): ert1q989xa8zz0a4fs7zrsmfsacwkvnup8vyd 989xa8zz0a4fs7zrsmfsacwkvnup8vydul8w9d
```` 
But a checksum of 38 chars for WitnessPubKeyHash not sure, that makes sense. 

So isn't there too much data in the p.WitnessHash? It looks like it.

I have been able to recompute the address I was expecting with the right checksum, by using bech32.ConvertBits on p.WitnessHash then cutting the results to keep the 32 first bytes. 

So that kind of comforts me into that idea that there are too much data in WitnessHash, or (most likely) just something out of my knowledge. 

Sorry the lengthy message, thanks for your help

- Sonny

### LINKED ISSUE #187: [payment] Fix WitnessHash set by FromScript for P2WPKH scripts
Refers to #186.

The problem is that [`payment.FromScript()`](../blob/master/payment/payment.go#L153) appends the pubkey hash of the witness program twice in its WitnessHash field.

altafan:
Closes #187.
Closes #186.

Please @tiero review this.

@SonnyAD you may want to test this to ensure the bug is fixed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,11 +20,8 @@ jobs:
       - name: Check out code into the Go module directory
         uses: actions/checkout@v2
 
-      - name: Install Nigiri
-        run: | 
-          mkdir ~/.nigiri; cd ~/.nigiri
-          curl https://travis.nigiri.network | bash; cd
-          docker-compose -f ~/.nigiri/docker-compose.yml up -d
+      - name: Run Nigiri
+        uses: vulpemventures/nigiri-github-action@v1
           
       - name: Get dependencies
         run: go get -v -t -d ./...
--- a/payment/payment.go
+++ b/payment/payment.go
@@ -127,11 +127,11 @@ func FromPayment(payment *Payment) (*Payment, error) {
 
 // FromScript creates parses a script into a Payment struct
 func FromScript(
-	script []byte,
+	outputScript []byte,
 	net *network.Network,
 	blindingKey *btcec.PublicKey,
 ) (*Payment, error) {
-	if script == nil || len(script) == 0 {
+	if len(outputScript) == 0 {
 		return nil, errors.New("payment's script can't be empty or nil")
 	}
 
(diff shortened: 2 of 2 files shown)


SonnyAD:
I confirm it works good for me. Thanks, that was quick.

PR #191, opened 2022-05-12
Output descriptors: `wpkh` fragment support

sekulicd:
This adds descriptor pkg with initial implementation for Wpkh descriptor wallet.

@tiero @altafan @louisinger please review.


descriptor/wpkh.go
```diff
@@ -0,0 +1,181 @@
+package descriptor
+
+import (
+	"encoding/hex"
+	"errors"
+
+	"github.com/btcsuite/btcutil/hdkeychain"
+
+	"github.com/btcsuite/btcd/txscript"
```
tiero, line 9: I am afraid that for elements we should have a way to extends it since elements has more opcodes. For now it's fine tho

descriptor/wpkh.go
```diff
@@ -0,0 +1,181 @@
...
+	"github.com/btcsuite/btcutil"
```
tiero, line 10: shouldn't be `github.com/btcsuite/btcd/btcutil` now?

sekulicd: we use github.com/btcsuite/btcutil in other pkg's, maybe we can treat this in another issue?

tiero:
utACK for me, I would wait an [actual implementation ](https://github.com/vulpemventures/neutrino-elements/issues/6)using this pkg before merging.

You can test there requiring this commit directly

PR #193, opened 2022-05-13
Wpkh descriptor parsing bug

sekulicd:
This fix parsing of wpkh descriptor in case fingerprint is not provided.

@tiero please review.


descriptor/parser_test.go
```diff
@@ -350,6 +350,27 @@ func TestParseScriptExpressionWpkh(t *testing.T) {
...
+				}
+
+				if hex.EncodeToString(scripts[0].Script) != scriptHexExpected {
+					return errors.New("unexpected script gen")
+				}
+
+				return nil
+			},
+			wantErr: false,
+		},
```
tiero, line 373: let's add also a test without child keys? ie. only the child xpub passes

sekulicd: Done

PR #194, opened 2022-05-26
Zero asset issuance

### LINKED ISSUE #190: addIssuance does not allow to make issuance of asset with 0 supply (and token > 0)
This is possible in Elements ie. https://blockstream.info/liquidtestnet/tx/ca1f6523efc7bd31208dbe6c8e2d675c6fd99ad07814da4358a20575aeb283a2?expand

sekulicd:
This adds fix for issuing zero value asset issuance.
Both blinded and unblinded zero value assets are supported.
Changes includes removing asset output if asset amount is zero and logic for calculating minValue for rangeProof which is decided based on scriptPubKey, check in [here](https://github.com/ElementsProject/elements/blob/80862ff9434fbb8f89bc189e92bb5af05db6144d/src/blind.cpp#L261)

This closes #190 


confidential/confidential.go
```diff
@@ -107,16 +113,21 @@ type RangeProofArgs struct {
...
 func (a RangeProofArgs) minValue() uint64 {
-	if a.MinValue <= 0 {
-		return 1
+	if isUnSpendable(a.ScriptPubkey) {
+		return 0
 	}
-	return a.MinValue
+
+	return 1
+}
```
altafan, line 126: did you take this from elements core codebase?

transaction/issuance.go
```diff
@@ -125,6 +125,11 @@ func NewTxIssuance(
...
 		return nil, err
 	}
+
+	if assetAmount == 0 {
+		confAssetAmount = []byte{0}
+	}
+
 	confTokenAmount, err := toConfidentialTokenAmount(
 		tokenAmount,
 	)
```
tiero, line 135: Couldnt happen to have in some cases also the token amount = 0? Ie. an Issuance without reissuance token (ofc in that case asset amount MUST be > 0 otherwise makes no sense)

transaction/issuance.go
```diff
@@ -125,6 +125,11 @@ func NewTxIssuance(
 	if err != nil {
 		return nil, err
 	}
+
+	if assetAmount == 0 {
+		confAssetAmount = []byte{0}
+	}
+
```
altafan, line 132: instead of adding this here, we could rename the actual `toConfidentialTokenAmount()` to a more generic `toConfidentialIssuanceAmount()`, and replace with it any occurrence of `toConfidentialAssetAmount()` and `toConfidentialTokenAmount()`.

pset/blinder.go
```diff
@@ -215,7 +215,10 @@ func (b *Blinder) unblindInputsToIssuanceBlindingData() (
 				return nil, err
 			}
 
-			value, _ := elementsutil.ElementsToSatoshiValue(input.Issuance.AssetAmount)
+			var value uint64 = 0
+			if len(input.Issuance.AssetAmount) == 9 {
```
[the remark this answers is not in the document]
altafan: This should be ok like this.
Actually, the name of the function is misleading: `unblindInputsToIssuanceBlindingData` does not actually unblind anything. This method is called by Blinder.Blind() in order to construct so-called pseudo inputs for any input issuance.
At this point, we are at the very early stages of the blinding process, the input issuances are still to be blinded, therefore their amounts are still revealed, waiting for being replaced by commitments later in the process.

We can eventually rename this method to something more appropriate like `getIssuancesPseudoInputs()` or similar.

transaction/issuance.go
altafan: `toConfidentialTokenAmount()` already takes care of returning `[]byte(0x00)` in case the token amount is 0.
In fact, we can make it more generic so that we can use it also for asset amount ([ref to comment](https://github.com/vulpemventures/go-elements/pull/194#discussion_r882575499))

confidential/confidential.go
sekulicd: Yeah, check comment in PR decs.

altafan:
@sekulicd watch out! [AddIssuanceArgs.validate()](https://github.com/sekulicd/go-elements/blob/zero_asset_issuance/pset/updater.go#L411-L413) still expects the asset address to be mandatory, while now it should be like that only in case the asset amount is > 0 (just like we do for the token the lines below).
Then in `TestBroadcastUnblindedZeroAssetValueIssuanceTx` we can avoid to pass an [AssetAddress](https://github.com/sekulicd/go-elements/blob/zero_asset_issuance/pset/pset_test.go#L844).

transaction/issuance.go
sekulicd: Done

sekulicd:
> @sekulicd watch out! [AddIssuanceArgs.validate()](https://github.com/sekulicd/go-elements/blob/zero_asset_issuance/pset/updater.go#L411-L413) still expects the asset address to be mandatory, while now it should be like that only in case the asset amount is > 0 (just like we do for the token the lines below). Then in `TestBroadcastUnblindedZeroAssetValueIssuanceTx` we can avoid to pass an [AssetAddress](https://github.com/sekulicd/go-elements/blob/zero_asset_issuance/pset/pset_test.go#L844).

Done

altafan (approved):

tiero (approved):

PR #195, opened 2022-05-31
Taproot support

louisinger:
**This PR adds Taproot/Tapscript support for elements, greatly inspired by the work already done on btcd**: https://github.com/btcsuite/btcd/pull/1787

### 1. bump btcd to latest, drop btcutils

* In order to use the latest changes of `btcsuite/btcd` (including taproot PRs + schnorr support), go mod is pointing the latest commit hash. It seems to me that taproot changes has not been released yet (Am I wrong ?). 
* btcutils packages are now included in btcd, that's why some imports have been modified

### 2. Elements taproot

* `taproot` package extends the `btcd/txscript/taproot` file and reuses the same unit test file `taproot_test.go`. Lot of copy-paste there, the logic is the same on Elements. The major change is about BIP-341 hash tags (suffixed by "/elements" on Liquid).
* `payment` pkg implements several new functions in order to support P2TR payments (unconfidential/confidential).
* `blech32` supports segwit v1 encoding (bech32m) + BONUS: add unit tests from liquidjs-lib
* `transaction` pkg implements `HashForWitnessV1`.
* some changes in `address` pkg (getScriptType, toOutputScript etc...).
* e2e nigiri testing for taproot (keypath & tapscript).
* add the new elements tapscript opcodes constants.

> _About new opcodes:_ It is quite possible to use the new opcodes now. However, for better introspection support, I think we should extend the scriptBuilder from txscript with some elements-specific functions like `AddAssetHash` or `AddIndex` etc.

@altafan @tiero please review


altafan (commented):
Missing test vectors for changes in address.go?

payment/p2tr.go
```diff
@@ -0,0 +1,170 @@
...
+		},
+	}, nil
+}
+
+func (p *Payment) taprootBech32() (*address.Bech32, error) {
+	payload := &address.Bech32{
+		Prefix:  p.Network.Bech32,
+		Version: segwitVersion,
+	}
+	if p.Taproot.XOnlyTweakedKey != nil {
```
altafan, line 102: Before this, we must ensure that `p.Taproot != nil` otherwise we must return an error.
If the user attempts to generate a TR address from a non TR payments, we must prevent panics so to speak.

payment/p2tr.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+	}
+
+	if payload.Program == nil {
+		return nil, errors.New("unable to compute taproot's tweaked key from payment data")
+	}
+
+	if len(payload.Program) != 32 {
+		return nil, errors.New("taproot's tweaked key has wrong length")
+	}
```
altafan, line 128: Do we really need these post checks? In general, this lib's methods make sure that everything's ok before doing any operation "by design", instead of performing post checks. 

Is there a way to make these checks over the taproot data before doing the magic in order to make sure that at this point `payload.Program` is exactly a 32-bytes buffer?

payment/p2tr.go
```diff
@@ -0,0 +1,170 @@
...
+//a confidential blech32 p2wsh address
+func (p *Payment) ConfidentialTaprootAddress() (string, error) {
+	bechTaproot, err := p.taprootBech32()
+	if err != nil {
+		return "", err
+	}
+
+	if p.BlindingKey == nil {
+		return "", errors.New("blinding key is required to derive confidential address")
+	}
```
altafan, line 156: this should be done before calling `p.taprootBech32()`

payment/payment.go
```diff
@@ -267,7 +275,7 @@ func (p *Payment) ConfidentialWitnessPubKeyHash() (string, error) {
 	}
 	addr, err := address.ToBlech32(payload)
 	if err != nil {
-		return "", nil
+		return "", err
```
altafan, line 278: Nice catch! :)

transaction/transaction.go
```diff
@@ -639,6 +643,120 @@ func (tx *Transaction) HashForWitnessV0(inIndex int, prevoutScript []byte, value
...
 }
 
+// HashForWitnessV1 returns the sighash as described in BIP-340
+func (tx *Transaction) HashForWitnessV1(
+	inIndex int,
+	prevoutsScripts [][]byte,
+	prevoutsAssetValues []struct {
+		Asset []byte
+		Value []byte
+	},
```
altafan, line 653: could it make sense to use a list of prevouts here like `prevouts Output[]` instead of these args?

Otherwise, iIwould go for `prevoutsScripts`, `prevoutsAssets` and `prevoutsValues` ... not so common to have untyped structs as func args.

altafan:
> About new opcodes: It is quite possible to use the new opcodes now. However, for better introspection support, I think we should extend the scriptBuilder from txscript with some elements-specific functions like AddAssetHash or AddIndex etc.

Don't think this is in the scope of the PR. We should open a dedicated ticket and discuss it there.

payment/payment.go
tiero: shall we make a dedicated PR and push right away?

sekulicd (commented):
It would be nice if we can add test cases examples with more complex scripts(multipath) for learning/doc purpose.

payment/p2tr.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/vulpemventures/go-elements/taproot"
+)
+
+const (
+	segwitVersion = byte(0x01)
+)
+
+// TaprootPaymentData is included in Payment struct to store Taproot-related data
+type TaprootPaymentData struct {
+	XOnlyTweakedKey    []byte
```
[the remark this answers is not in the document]
sekulicd: From my POV user should be able to access fields but not to mutate them(which is done with factory methods), this way we make sure pkg is not misused.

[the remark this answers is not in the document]
altafan: factories, as the name suggests, are intended to be constructors for exported/unexported types.

The manipulation of the fields can be done via type's exported methods, or directly by the user if the inner fields are exported.

This is a library and its purpose is to give the user the building blocks to create any kind of application.
For this reason, I do think it's better to allow the user to make mistakes instead of him not being able to use the lib because it misses methods to do some very fancy stuff.

After all, all the rest of the library is designed to achieve this goal and there doesn't seem to be any valid reason to change it starting from now.

taproot/taproot_e2e_test.go
```diff
@@ -0,0 +1,436 @@
...
+
+	unsignedTx.Inputs[0].Witness = transaction.TxWitness{
+		sig.Serialize(),
+	}
+
+	signed, err := unsignedTx.ToHex()
+	if err != nil {
+		t.Fatal(err)
+	}
+
```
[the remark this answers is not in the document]
louisinger: Not necessery in my opinion. VerifyTaprootLeafCommitment is tested in taproot pkg and broadcast is already a "sig check"

payment/payment.go
louisinger: #196

PR #202, opened 2022-06-23
Generate address from output script

sekulicd:
This adds FromOutputScript func to the payment pkg which is supposed to create address from provided output script.


.golangci.yml
```diff
@@ -0,0 +1,57 @@
+run:
```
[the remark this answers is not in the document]
sekulicd: sorry, this is some leftover, removed it.

payment/payment.go
```diff
@@ -373,3 +373,89 @@ func buildScript(hash []byte, scriptType string) []byte {
 	script, _ := builder.Script()
 	return script
 }
+
+func FromOutputScript(
```
altafan, line 377: This method should stay in `address` package and should not make use of the payment one, like a "raw" address builder.
The idea is to allow the user to retrieve and address from an output script with either `address` or `payment`. The former lacks such API, while the latter already provides the necessary ones.

sekulicd: yeah, thing is that if adding it to address pkg can use payment cause of cyclic import.

altafan: I was trying to say that this function should be moved to `address` and its implementation must change so that it doesn't make use of `payment`'s types/functions.

altafan: I suggest you don't look at the liquidjs-lib implementation, rather this function should analyze the incoming script to understand its type and encode it to an address using the proper format (base58, bech32...). 
I would remove also the blinding key argument and return always an unconfidential address from an output script to keep it simple. The user can translate it to confidential with existing types and functions of the `address` package if needed.

PR #211, opened 2022-10-26
Pset2 new fields

### LINKED ISSUE #208: [psetv2] Add new partial input fields
The [spec](https://github.com/ElementsProject/elements/blob/master/doc/pset.mediawiki#specification) now defines 3 new per-input proprietary fields that must be added to our implementation:
 - `PSBT_ELEMENTS_IN_EXPLICIT_ASSET`
 - `PSBT_ELEMENTS_IN_ASSET_PROOF`
 - `PSBT_ELEMENTS_IN_BLINDED_ISSUANCE`

sekulicd:
This adds bellow PSET missing fields:
PSBT_ELEMENTS_IN_EXPLICIT_VALUE
PSBT_ELEMENTS_IN_VALUE_PROOF
PSBT_ELEMENTS_IN_EXPLICIT_ASSET
PSBT_ELEMENTS_IN_ASSET_PROOF
PSBT_ELEMENTS_IN_BLINDED_ISSUANCE

This closes #208 


psetv2/input.go
```diff
@@ -690,6 +712,66 @@ func (i *Input) getKeyPairs() ([]KeyPair, error) {
...
+	if i.BlindedIssuance {
+		blindedIssuance = []byte{1}
+	}
+	assetProofKeyPair := KeyPair{
+		Key: Key{
+			KeyType: PsetProprietary,
+			KeyData: proprietaryKey(InputBlindedIssuanceValue, nil),
+		},
+		Value: blindedIssuance,
+	}
```
altafan, line 772: The spec says:
> A boolean flag. 0x00 indicates the issuance should not be blinded, 0x01 indicates it should be. If not specified, assumed to be 0x01. Note that this does not indicate actual blinding status, but rather the expected blinding status prior to signing.

To correctly follow the spec we have to declare `BlindedIssuance` as `*bool`. This way, in case it's not defined we can default to 0x01.

altafan: Also, with this kind of declaration we shouldn't break unit tests because in case the field is nil we just don't serialize it.

sekulicd: Done

psetv2/input.go
```diff
@@ -1034,6 +1118,38 @@ func (i *Input) deserialize(buf *bytes.Buffer) error {
...
+						return ErrInDuplicatedField("value proof")
+					}
+					i.ValueProof = kp.Value
+				case InputExplicitAsset:
+					if len(i.ExplicitAsset) > 0 {
+						return ErrInDuplicatedField("explicit asset")
+					}
+					if len(kp.Value) != 32 {
+						return ErrInInvalidExplicitAsset
+					}
```
altafan, line 1140: Missing assignment:
```suggestion
					if len(i.ExplicitAsset) > 0 {
						return ErrInDuplicatedField("explicit asset")
					}
					if len(kp.Value) != 32 {
						return ErrInInvalidExplicitAsset
					}
					i.ExplicitAsset = kp.value
```

psetv2/input.go
```diff
@@ -1034,6 +1118,38 @@ func (i *Input) deserialize(buf *bytes.Buffer) error {
...
+				case InputAssetProof:
+					if len(i.AssetProof) > 0 {
+						return ErrInDuplicatedField("asset proof")
+					}
+					i.AssetProof = kp.Value
+				case InputBlindedIssuanceValue:
+					if len(kp.Value) != 1 {
+						return ErrInInvalidBlindedIssuanceValue
+					}
```
altafan, line 1149: Missing check for duplicated field:
```suggestion
					if i.BlindedIssuance != nil {
					        return ErrInDuplicatedField("blinded issuance flag")
					}
					
					if len(kp.Value) != 1 {
						return ErrInInvalidBlindedIssuanceValue
					}
```

sekulicd: Done

psetv2/input.go
sekulicd: Done

psetv2/input.go
```diff
@@ -1148,6 +1151,10 @@ func (i *Input) deserialize(buf *bytes.Buffer) error {
 						return ErrInInvalidBlindedIssuanceValue
 					}
 
+					if i.BlindedIssuance != nil {
+						return ErrInDuplicatedField("blinded issuance flag")
+					}
```
altafan, line 1156: in the rest of the code we first check for duplicates,  then for value correctness. Can you please move this before the value validation if block?

sekulicd: Done

altafan (approved):

PR #230, opened 2024-10-16
Add support for Discount CT

michael1011:
Adds support for [Discount CT](https://github.com/ElementsProject/ELIPs/blob/main/elip-0200.mediawiki)

Test fixtures are from the ELIP


altafan:
@michael1011 thank you for the contribution!

